PR #13, opened 2019-02-13
Allow setting case on Xrefs

PiotrKozlowski:
@aiwenar Edit: Handled. ~~`xmlnsCmlnle="http://katalysteducation.org/cmlnle/1.0"` is transformed to `xmlns-cmlnle="http://katalysteducation.org/cmlnle/1.0"`~~


src/cnxml/deserialize.js
```diff
@@ -177,13 +177,14 @@ function emphasis(el) {
...
     const url = el.getAttribute('url') || null
+    const declension = el.getAttribute('cmlnle-case') || null
 
     if (target) {
         return {
             object: 'inline',
             type: 'xref',
             isVoid: true,
-            data: { target },
+            data: { target, declension },
```
aiwenar: This should be called _case_; _declension_ in English is not the same as a _grammatical case_.

> **declension**
>
> 1. A falling off, decay or descent.
> 2. (grammar) The act of declining a word; the act of listing the inflections of a noun, pronoun or adjective in order.
> 3. (grammar) A way of categorizing nouns, pronouns, or adjectives according to the inflections they receive. 
>
> (from [Wiktionary](https://en.wiktionary.org/wiki/declension))

> **(grammatical) case**
>
> _[..]_
>
> 6. (grammar) A specific inflection of a word depending on its function in the sentence.
> 
> and
>
> 1. (grammar) A mode of inflection of a word dependent on its use, especially the syntactic function in a phrase. 
>
> (from [Wiktionary](https://en.wiktionary.org/wiki/case#Noun), [Wiktionary](https://en.wiktionary.org/wiki/grammatical_case#English))

src/cnxml/serialize.js
```diff
@@ -75,6 +75,14 @@ const BLOCK_TAGS = {
 }
 
 
+/**
+ * List of valid declensions for xrefs.
+ * 
+ * @see xref
+ */
+const DECLENSIONS = ['genitive', 'dative', 'accusative', 'instrumental', 'locative', 'vocative']
```
aiwenar: You shouldn't limit it to just cases used in Polish. If you want a list of all existing grammatical cases (at least according to Wikipedia) I should still have it somewhere.

src/cnxml/serialize.js
```diff
@@ -132,7 +140,22 @@ function list(obj, children) {
...
+    let attrs = {
+        'target-id': obj.data.get('target')
+    }
+
+    const declension = obj.data.get('declension')
+    const cmlnleCase = declension ?
+        DECLENSIONS.some(dec => dec === declension) ?
+            declension
+            : 'INVALID'
+        : null
```
aiwenar: Why do you have this check here? You are guaranteed that the document you are serializing is correct according to its schema, and if you want to limit values of `cmlnle:case` then you should do it in schema.

src/plugins/xref/schema.js
```diff
@@ -17,7 +17,8 @@ export default {
             // TODO: better data validation
             data: {
                 document: value => value == null || typeof value === 'string',
-                target: Boolean,
+                target: t => t === null || typeof t === 'string',
+                declension: d => d === null || d === undefined || typeof d === 'string',
```
aiwenar: 1. `target` must be defined, so this check is wrong
2. rather than testing for `null` and `undefined` separately, check just for `d == null`

src/plugins/xref/queries.js
```diff
@@ -0,0 +1,13 @@
...
+
+  return inline.type === 'xref' ? inline : null
+}
+
+export function getActiveLink(editor, value) {
+  const inline = value.startInline
+  if (!inline) return null
+
+  return inline.type === 'link' ? inline : null
+}
```
aiwenar: Given how trivial those queries are, I'd rather not have them as separate functions.

src/plugins/xref/schema.js
PiotrKozlowski: Not all xrefs will have `data.declension` so `undefined` is needes as well as `null`

src/cnxml/serialize.js
PiotrKozlowski: Sure, please send me if you find it or feel free to add it

PiotrKozlowski: But wait, is cnxml supporting all those cases?

src/cnxml/serialize.js
PiotrKozlowski: I was thinking this way:
schema should allow any string as `case`, but when serializing to cnxml we have to be sure that it is supported.

src/plugins/xref/schema.js
PiotrKozlowski: `target: typeof t === 'string'` will be ok then?

src/cnxml/serialize.js
aiwenar: This is our extension to CNXML, and there isn't even any document documenting it :) But the idea was that you can set any valid case.

src/cnxml/serialize.js
aiwenar: This makes no sense. Why would you allow invalid values anywhere?

src/plugins/xref/schema.js
aiwenar: 1. Note that I proposed `d == null`, with only two equals signs. This is the same as checking with three equal signs for `d === null || d === undefined`

2. Technically in Slate keys need not be strings (or at least I think so), that's why I was only checking for their presence and not type

src/cnxml/serialize.js
PiotrKozlowski: In case when we want to serialize to another format which may be supporting those cases

src/plugins/xref/schema.js
PiotrKozlowski: `d === null || d === undefined` - this is exactly why I've like to separate those checks, but it's your repo so I'll adjust

aiwenar: In general I agree that it's better to use explicit checks, but here we have so many checks that it just becomes hard to read.

src/cnxml/deserialize.js
```diff
@@ -177,13 +177,14 @@ function emphasis(el) {
 function xref(el) {
     const target = el.getAttribute('target-id') || null
     const url = el.getAttribute('url') || null
+    const cmlnleCase = el.getAttribute('cmlnle-case') || null
```
aiwenar: This won't work, you need to use [`getAttributeNS`](https://developer.mozilla.org/en-US/docs/Web/API/Element/getAttributeNS),

src/plugins/xref/schema.js
```diff
@@ -18,6 +26,7 @@ export default {
             data: {
                 document: value => value == null || typeof value === 'string',
                 target: Boolean,
+                case: c => c == null || DECLENSIONS.some(d => d === c),
```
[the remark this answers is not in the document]
PiotrKozlowski: :man_facepalming: I was lately using `some()` ins other project and I've totally forgot about this

src/cnxml/deserialize.js
PiotrKozlowski: There was type `-` instead of `:`
`getAttributeNS('cmlnle', 'case)` was not working but `getAttribute('cmlnle:case')` is ok.

aiwenar: `getAttribute('cmlnle:case')` also doesn't work:

```js
const document = `<?xml version="1.0" encoding="utf-8"?>
<document xmlns="http://cnx.rice.edu/cnxml" xmlns:cmlnle="http://katalysteducation.org/cmlnle/1.0" xmlns:n1="http://katalysteducation.org/cmlnle/1.0">
    <test cmlnle:case="nominative" />
    <test n1:case="accusative" />
</document>`

const tree = new DOMParser().parseFromString(d, 'application/xml')

console.log(t.children[0].children[0].getAttribute('cmlnle:case'))
// => nominative
console.log(t.children[0].children[1].getAttribute('cmlnle:case'))
// => null
```

The reason that `getAttributeNS` didn't work for you is because you used it wrong. `cmlnle` is _not_ a namespace name, its a namespace _prefix_, which the user is free to choose (as I did here, using both `cmlnle` and `n1` to refer to the same namespace). The namespace is actually `http://katalysteducation.org/cmlnle/1.0`. Continuing with example:

```js
console.log(t.children[0].children[0].getAttributeNS('http://katalysteducation.org/cmlnle/1.0', 'case'))
// => nominative
console.log(t.children[0].children[1].getAttributeNS('http://katalysteducation.org/cmlnle/1.0', 'case'))
// => accusative
```

aiwenar:
I've just realised that transformation from `nsprefixAttrname` to `nsprefix:attrname` is not part of React or JSX but was implemented by me in [src/cnxml/xml.js](https://github.com/katalysteducation/cnx-designer/blob/master/src/cnxml/xml.js). All you need to do to make it work is add `cmlnle: 'http://katalysteducation.org/cmlnle/1.0',` to `NAMESPACES` in that file.

src/cnxml/deserialize.js
PiotrKozlowski: Strange, I've tested on live editor and `getAttribute('cmlnle:case')` is working.

aiwenar: Apparently it does when you use the same namespace prefix as is used in the document, but it won't work if you use a different one, as I've shown.

PiotrKozlowski:
@aiwenar Should we add something more here?

src/cnxml/index.js
```diff
@@ -30,7 +30,7 @@ const serializer = new Html({
 export default {
     deserialize: (...args) => serializer.deserialize(...args),
 
-    serialize(value) {
+    serialize(value) {
```
aiwenar: In the future strip such changes before sending a PR.

aiwenar:
Also when possible try to make PR history linear, having multiple merge commits within a feature branch makes it hard to follow.

src/cnxml/index.js
PiotrKozlowski: How can I do this?

PR #14, opened 2019-02-15
[WIP] Tables

PiotrKozlowski:
Two issues:
- I'm losing all attributes from elements in either serialization or deserialization.
- How to make table cells editable but do not allow to delete them?

Styling for tables: https://github.com/katalysteducation/adaptarr-front/pull/19

### DIFF
--- a/src/plugins/table/schema.js
+++ b/src/plugins/table/schema.js
@@ -1,0 +1,182 @@
+// Copyright 2018 OpenStax Poland
+// Licensed under the MIT license. See LICENSE file in the project root for
+// full license text.
+
+function normalizeTable(change, error) {
+    const { code: violation, child } = error
+
+    switch (violation) {
+        case 'child_type_invalid':
+            console.log('child_type_invalid')
+            console.log('child.type:', child.type)
+            if (child.type === 'title') {
+                change.setNodeByKey(child.key, { type: 'caption' })
+            } else {
+                console.warn('Unhandled table violation:', violation, JSON.stringify(error, null, 2))
+            }
+            break
+
+        default:
+            console.warn('Unhandled table violation:', violation)
+            break
+    }
+}
+
+function normalizeTgroup(change, error) {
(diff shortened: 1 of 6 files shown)


aiwenar:
Could you rebase this on top of master? GitHub seems confused about what's part of this PR and what's not. And in the future, please don't base feature branches on other feature branches.

PiotrKozlowski:
@aiwenar Done

src/plugins/table/schema.js
```diff
@@ -0,0 +1,182 @@
...
+export default {
+    blocks: {
+        table: {
+            data: {
+                summary: s => s == null || typeof s === 'string',
+                frame: f => f == null || typeof f === 'string',
+                colsep: c => c == null || c >= 0,
+                rowsep: r => r == null || r >= 0,
+                pgwide: p => p == null || p >= 0,
+            },
```
aiwenar: Even though it's a temporary solution we might still want a proper schema for those attributes.

src/plugins/table/schema.js
```diff
@@ -0,0 +1,182 @@
...
+            nodes: [
+                { match: { type: 'caption' }, min: 0, max: 1, },
+                { match: { type: 'tgroup' }, min: 1, },
+                { match: { type: 'caption' }, min: 0, max: 1, },
+            ],
+            normalize: normalizeTable,
+        },
+        tgroup: {
+            data: {
+                cols: c => c == null || c > 0,
```
aiwenar: `cols` is always required.

aiwenar:
I've also just realised, that rather than adding this temporarily to `cnx-designer`, we could keep it as a plugin in `adaptarr-front`. This would require some changes to CNXML de/serialization to allow for defining custom rules in plugins though.

PiotrKozlowski:
I've opened separate. Now it's handled as separate plugin on the front end - https://github.com/katalysteducation/cnx-designer/pull/19

PR #15, opened 2019-02-19
Handle classes

PiotrKozlowski:
What elements should support classes?
From one perspective we could handle all elements, but in another we probably will never use class on paragraphs or list items.

https://github.com/katalysteducation/adaptarr-front/pull/20

### DIFF
--- a/src/cnxml/deserialize.js
+++ b/src/cnxml/deserialize.js
@@ -2,6 +2,8 @@
 // Licensed under the MIT license. See LICENSE file in the project root for
 // full license text.
 
+import { List } from 'immutable'
+
 /**
  * Load a block element.
  *
@@ -14,6 +16,9 @@ const BLOCK = {
 
         const props = block instanceof Function ? block(el, next) : {
             type: block,
+            data: {
+                class: loadClasses(el),
+            },
             nodes: next(Array.from(el.children)),
         }
 
@@ -83,6 +88,9 @@ const text = type => (el, next) => ({
  */
 const mixed = type => (el, next) => ({
     type: type,
+    data: {
+        class: loadClasses(el),
+    },
     nodes: mixedContent(el, next),
 })
 
@@ -149,6 +157,7 @@ function admonition(el, next) {
         key: el.getAttribute('id') || undefined,
         data: {
             type: el.getAttribute('type') || 'note',
+            class: loadClasses(el),
         },
         nodes: mixedContent(el, next),
     }
@@ -204,6 +213,9 @@ function xref(el) {
 function list(el, next) {
     return {
         type: el.getAttribute('type') === 'enumerated' ? 'ol_list' : 'ul_list',
+        data: {
+            class: loadClasses(el),
+        },
         nodes: next(Array.from(el.children)),
     }
 }
@@ -265,6 +277,18 @@ function mixedContent(el, next, type='paragraph') {
     } else {
         return next(Array.from(el.children))
     }
+}
+
+
+/**
+ * Return element's classes as an array.
+ */
+const loadClasses = (el) => {
+    const classes = el.getAttribute('class')
+
+    if (!classes) return List([])
+
+    return List(classes.trim().split(/\s+/))
 }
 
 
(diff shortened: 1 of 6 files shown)


src/cnxml/serialize.js
```diff
@@ -106,7 +113,9 @@ const MARK_TAGS = {
  * Serializer for figures.
  */
 function figure(obj, children) {
-    return <figure id={obj.key}>
+    let attrs = obj.data.get('class') ? {class: obj.data.get('class') } : {}
```
aiwenar: Is there any reason why you are using `obj.data.toJS()` for some nodes, and manually constructing it for others? Also, `obj.data.has()`.

PiotrKozlowski: `obj.data.toJS()` is passing all arguments, even empty ones and I didn't like that, but probably better way of handling this would be `<figure id={obj.key} ${...removeEmptyKeys(obj.data.toJS())}>`
Would you like me to implement this solution?

src/plugins/admonition/schema.js
```diff
@@ -36,7 +36,7 @@ export default {
         admonition: {
             data: {
                 type: v => TYPES.includes(v),
-                class: c => c === undefined || typeof c === 'string',
+                class: c => c == null || Array.isArray(c),
```
aiwenar: As I said, I would prefer it to be an `Immutable.List`. And since we keep classes in a list/array, we should also check that there are no spaces in classes, 'c.every(x => x.match(/\s/) == null)` should work both for `List` and for `Array`.

src/cnxml/deserialize.js
```diff
@@ -77,7 +77,7 @@ const INLINE_TAGS = [
 const text = type => (el, next) => ({
     type: type,
     data: {
-        class: el.getAttribute('class') || '',
+        class: loadClasses(el),
     },
```
aiwenar: We don't support classes on text elements. And while loading them is not a big issue, as they will be removed by schema, there's also no reason to do so.

src/cnxml/deserialize.js
```diff
@@ -279,7 +274,7 @@ function mixedContent(el, next, type='paragraph') {
             object: 'block',
             type: type,
             data: {
-                class: el.getAttribute('class') || '',
+                class: loadClasses(el),
             },
```
aiwenar: This block is actually a child of `el` and should't receive its classes.

src/plugins/admonition/schema.js
PiotrKozlowski: I've changed it to Immutable List.
I'm already handling spaces in class names in front and i'm not sure about putting this into schema validation because it will remove all classes even if only one of them is wrong.

src/cnxml/serialize.js
```diff
@@ -22,7 +22,7 @@ const BLOCK = {
         if (!data.class) {
             // Remove empty classes
             delete data['class']
-        } else {
+        } else {node
```
aiwenar: What's this?

src/cnxml/deserialize.js
```diff
@@ -2,6 +2,8 @@
 // Licensed under the MIT license. See LICENSE file in the project root for
 // full license text.
 
+import Immutable from 'immutable'
```
aiwenar: Why not `import { List } from 'immutable'`?

src/plugins/admonition/schema.js
```diff
@@ -36,7 +38,7 @@ export default {
         admonition: {
             data: {
                 type: v => TYPES.includes(v),
-                class: c => c == null || Array.isArray(c),
+                class: c => c == null || Immutable.List.isList(c),
```
aiwenar: It would be good to also check whether classes themself are correct (i.e. they don't contain white spaces).

src/cnxml/deserialize.js
aiwenar: 1. What makes titles different from other types of text (paragraphs) that they get to have classes?

2. Could you please stop marking conversations as resolved? It is _not your call_ to say that _my_ concerns were addressed.

PR #16, opened 2019-02-20
Improved schemas, tests, commands, and queries for sections and titles

aiwenar:


src/plugins/section/commands.js
```diff
@@ -0,0 +1,126 @@
...
+    const { value } = change
+    const section = change.getActiveSubsection(value)
+
+    // No section is selected
+    if (section == null) return
+
+    const parent = value.document.getParent(section.key)
+    const index = parent.nodes.indexOf(section)
+    const pparent = value.document.getParent(parent.key)
+    const pindex = pparent.nodes.indexOf(parent)
```
PiotrKozlowski: `pparent` seems to be `null` when decreasing section depth for first level section.

aiwenar: Right, nicely spotted!

PR #17, opened 2019-03-04
Support for inline and block <code> tags

PiotrKozlowski:
UI: https://github.com/katalysteducation/adaptarr-front/pull/25

### DIFF
--- a/src/cnxml/deserialize.js
+++ b/src/cnxml/deserialize.js
@@ -107,6 +107,7 @@ const mixed = type => (el, next) => ({
  */
 const BLOCK_TAGS = {
     caption: text('figure_caption'),
+    code: code,
     commentary: mixed('exercise_commentary'),
     exercise: 'exercise',
     figure: 'figure',
@@ -151,6 +152,20 @@ function admonition(el, next) {
             type: el.getAttribute('type') || 'note',
         },
         nodes: mixedContent(el, next),
+    }
+}
+
+
+/**
+ * Process data for code tags.
+ */
+function code(el, next) {
+    const display = el.getAttribute('display') || 'inline'
+
+    return {
+        object: display === 'block' ? 'block' : 'inline',
+        type: 'code',
+        nodes: next(el.childNodes),
     }
 }
 
--- a/src/cnxml/serialize.js
+++ b/src/cnxml/serialize.js
@@ -58,6 +58,7 @@ const MARK = {
  */
 const BLOCK_TAGS = {
     admonition: 'note',
+    code: code,
     exercise: 'exercise',
     exercise_commentary: 'commentary',
     exercise_problem: 'problem',
@@ -145,6 +146,22 @@ function xref(obj, children) {
     return <link {...attrs}>
         {children}
     </link>
+}
+
+
+/**
+ * Serializer for code.
+ */
+function code(obj, children) {
+    let attrs = {}
+
+    if (obj.object === 'block') {
+        attrs.display = 'block'
+    }
+
+    return <code {...attrs}>
+        {obj.getTexts().map(t => t.text)}
+    </code>
 }
 
 
--- a/src/plugins/section/schema.js
+++ b/src/plugins/section/schema.js
@@ -73,6 +73,7 @@ function normalizeSection(change, error) {
  */
 const DOCUMENT_NODES = [
     { type: 'admonition' },
+    { type: 'code', },
     { type: 'exercise' },
     { type: 'figure' },
     { type: 'ol_list' },
--- a/src/plugins/text/render.js
+++ b/src/plugins/text/render.js
@@ -9,11 +9,11 @@ const NODES = {
 }
 
 const MARKS = {
-    strong: 'strong',
     emphasis: 'em',
     underline: 'u',
     superscript: 'sup',
     subscript: 'sub',
+    strong: 'strong',
 }
 
 export function renderNode({ node, children, attributes }, editor, next) {
--- a/src/plugins/text/schema.js
+++ b/src/plugins/text/schema.js
@@ -6,11 +6,11 @@ export default {
     blocks: {
         paragraph: {
             marks: [
-                { type: 'strong' },
                 { type: 'emphasis' },
                 { type: 'underline' },
                 { type: 'superscript' },
                 { type: 'subscript' },
+                { type: 'strong' },
             ],
         },
     },
--- a/test/plugins/code/handle-backspace.js
+++ b/test/plugins/code/handle-backspace.js
@@ -1,0 +1,20 @@
+/** @jsx h */
+
+export default (change, editor) => {
+  editor.run('onKeyDown', { key: 'Backspace' })
+}
+
+export const input = <value>
+    <document>
+        <p>Before</p>
+        <code><cursor/>Some code</code>
+        <p>After</p>
+    </document>
+</value>
+
+export const output = <value>
+    <document>
+        <p>Before<cursor/>Some code</p>
+        <p>After</p>
+    </document>
+</value>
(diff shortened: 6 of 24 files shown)


src/plugins/code/queries.js
```diff
@@ -0,0 +1,10 @@
+// Copyright 2018 OpenStax Poland
+// Licensed under the MIT license. See LICENSE file in the project root for
+// full license text.
+
+export function getActiveCode(editor, value) {
+  const block = value.startBlock
+  if (!block) return null
+
+  return  block.type === 'code' ? block : null
```
aiwenar: I think `getActiveSth` queries should only be created for blocks where this operations is not trivial. Admonitions, exercises, lists, and sections have them because selection may be arbitrarily deep inside them, and so finding whether you are inside one of them is not trivial (which just made me realise that `getActiveExercise` and `getActiveAdmonition` are broken as they assume at most one level of nesting, but can contain lists...). Here isn't any such reason.

src/cnxml/deserialize.js
```diff
@@ -121,6 +121,7 @@ const BLOCK_TAGS = {
     solution: 'exercise_solution',
     subfigure: 'figure',
     title: text('title'),
+    code: text('code'),
```
aiwenar: Properties in this object should be sorted alphabetically.

src/cnxml/deserialize.js
```diff
@@ -137,6 +138,7 @@ const MARK_TAGS = {
     sub: 'subscript',
     sup: 'superscript',
     link: xref,
+    code: 'code',
 }
```
aiwenar: Properties in this object should be sorted alphabetically (yep I forgot about it when adding `link`).

src/cnxml/serialize.js
```diff
@@ -72,6 +72,7 @@ const BLOCK_TAGS = {
     section: 'section',
     title: 'title',
     ul_list: list,
+    code: code,
 }
```
aiwenar: Properties in this object should be sorted alphabetically.

src/cnxml/serialize.js
```diff
@@ -99,6 +100,7 @@ const MARK_TAGS = {
     superscript: 'sup',
     underline: emphasis('underline'),
     xref: xref,
+    code: 'code',
 }
```
aiwenar: Properties in this object should be sorted alphabetically.

src/cnxml/serialize.js
```diff
@@ -148,6 +150,16 @@ function xref(obj, children) {
...
 
 
+/**
+ * Serializer for code.
+ */
+function code(obj, children) {
+    return <code display="block">
+        {obj.getTexts().map(t => t.text)}
+    </code>
```
aiwenar: Why are you manually serializing contents rather than using `children`?

src/plugins/section/schema.js
```diff
@@ -15,12 +15,19 @@ function normalizeDocument(change, error) {
...
-            change.setNodeByKey(child.key, { type: 'paragraph' })
-            return
+        switch (child.type) {
+            case 'title':
+                change.setNodeByKey(child.key, { type: 'paragraph' })
+                return
+
+            case 'code_line':
+                change.setNodeByKey(child.key, { type: 'paragraph' })
+                return
```
aiwenar: What is `'code_line'`? You neither define nor use such node.

src/plugins/text/schema.js
```diff
@@ -11,6 +11,7 @@ export default {
                 { type: 'underline' },
                 { type: 'superscript' },
                 { type: 'subscript' },
+                { type: 'code' },
```
aiwenar: I think `code` should be an inline. Marks can overlap, so you could end up with e.g. an `emphasis` inside a `code` span. Inlines can (if I'm not mistaken) specify what marks they allow inside them.

src/plugins/text/render.js
```diff
@@ -14,6 +14,7 @@ const MARKS = {
     underline: 'u',
     superscript: 'sup',
     subscript: 'sub',
+    code: 'code',
```
aiwenar: Plugins should only render nodes which they define, and plugins which define a node should also render it.

test/plugins/code/insert.js
```diff
@@ -0,0 +1,16 @@
+/** @jsx h */
+
+export default change => change.insertCode()
+
+export const input = <value>
+    <document>
+        <p><cursor/>Some code</p>
+    </document>
```
aiwenar: Move `<cursor/>` inside text to check that it's being split properly. Also add a test in which code is added into an expanded selection (I think that in such case selected text should be moved into new code block).

test/plugins/code/handle-backspace.js
```diff
@@ -0,0 +1,20 @@
...
+
+export default (change, editor) => {
+  editor.run('onKeyDown', { key: 'Backspace' })
+}
+
+export const input = <value>
+    <document>
+        <p>Before</p>
+        <code><cursor/>Some code</code>
+        <p>After</p>
```
aiwenar: Also add a variation of test in which code has multiple lines.

src/plugins/text/schema.js
PiotrKozlowski: Gutenberg handle this as a mark. Slate also provide example with `code` as a mark.

src/plugins/text/render.js
PiotrKozlowski: Not sure about this. If we have all others marks in one place then why change it?

src/plugins/text/schema.js
aiwenar: Gutenberg is Slate based? And as for Slate's example, see what happens when you add other marks inside that inline code snippet. I don't think such behaviour is acceptable.

test/plugins/code/handle-enter.js
```diff
@@ -16,6 +17,7 @@ export const output = <value>
     <document>
         <code>
             Some code
+            {'\n'}
```
aiwenar: Other elements don't leave an empty line when you exit them via enter.

src/plugins/text/schema.js
PiotrKozlowski: In both of this editor marks like `strong`, `emphasis` are allowed inside `code` mark.

src/plugins/text/render.js
aiwenar: Plugins are (supposed to be, by design) independent from each other. So if you disable plugin `Text`, `Code` should work without any change in behaviour, but as it is right now, inline code tags will stop rendering.

src/plugins/text/schema.js
aiwenar: Yes, and it looks terrible and makes no sense.

PiotrKozlowski: It makes sens because it's text editor and not code editor so you should always have some kind of way to make some fragment bold or italic, etc. 
It's also allowed according to the documentation.
However if you want me to make it inline then we have to first get `wrapInline()` function to work again.

PiotrKozlowski:
It works with https://github.com/katalysteducation/slate-counters/pull/3

src/cnxml/deserialize.js
```diff
@@ -155,6 +156,22 @@ function admonition(el, next) {
...
 
 
+/**
+ * Process data for emphasis marks.
+ */
+function code(el, next) {
+    let object = 'block'
+    if (el.parentNode && el.parentNode.tagName.toLowerCase() === 'para') {
```
aiwenar: Inline code fragments can be placed in more than just paragraphs (e.g. figure captions).

aiwenar: Since the `display` attribute is required for block code fragments, we could just case over `getAttribute('display') || 'inline'`.

PiotrKozlowski:
@aiwenar up

src/cnxml/serialize.js
PiotrKozlowski: `"element.props.children is undefined"`

PR #18, opened 2019-03-13
Initial support for suggestions in text

PiotrKozlowski:
front - https://github.com/katalysteducation/adaptarr-front/pull/26


aiwenar:
Was implemented in openstax-poland/adaptarr-front#48.

PR #19, opened 2019-03-15
Allow defining custom CNXML (de)serialization rules

PiotrKozlowski:
Front: https://github.com/katalysteducation/adaptarr-front/pull/27

I've realized that to allow user setting he's own [de]serialization rules we need only export those:
```
parseHtml,
deserializeRules: deserialize,
serializeRules: serialize,
render
```
So now on the front end we can use built int serializer or just create our own.

I've changed `caption` to be handled as `text('caption')` and the normalization should change it to `figure/table_caption`.

### DIFF
--- a/src/cnxml/deserialize.js
+++ b/src/cnxml/deserialize.js
@@ -148,7 +148,7 @@ const mixed = type => (el, next) => ({
  * @see BLOCK
  */
 const BLOCK_TAGS = {
-    caption: text('figure_caption'),
+    caption: caption,
     commentary: mixed('exercise_commentary'),
     exercise: 'exercise',
     figure: 'figure',
@@ -179,6 +179,23 @@ const MARK_TAGS = {
     sub: 'subscript',
     sup: 'superscript',
     link: xref,
+}
+
+
+/**
+ * Process data for captions.
+ */
+function caption(el, next) {
+    const allowedParents = ['figure', 'subfigure']
+
+    if (!allowedParents.includes(el.parentElement.tagName)) {
+        return
+    }
+
+    return splitBlocks({
+        type: 'figure_caption',
+        nodes: next(el.childNodes),
+    })
 }
 
 
--- a/src/cnxml/index.js
+++ b/src/cnxml/index.js
@@ -25,15 +25,18 @@ function parseHtml(html) {
 }
 
 
-const serializer = new Html({
-    rules: [...deserialize, ...serialize],
-    defaultBlock: 'invalid',
-    parseHtml,
-})
+export default class CNXML {
+    constructor(rules = []) {
+        this.serializer = new Html({
+            rules: [...rules, ...deserialize, ...serialize],
+            defaultBlock: 'invalid',
+            parseHtml,
+        })
+    }
 
-
-export default {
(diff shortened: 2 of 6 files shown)


src/cnxml/index.js
```diff
@@ -45,10 +52,13 @@ export default {
...
+                {this.serializer.serialize(value, { render: false })}
             </content>
         </document>
 
         return render(r, options)
     }
 }
+
+
+export default CNXML
```
aiwenar: Use `export default class`.

src/cnxml/index.js
```diff
@@ -25,15 +25,22 @@ function parseHtml(html) {
 }
 
 
-const serializer = new Html({
-    rules: [...deserialize, ...serialize],
-    defaultBlock: 'invalid',
-    parseHtml,
-})
+class CNXML {
+    customRules = []
```
aiwenar: What is this for?

src/cnxml/deserialize.js
```diff
@@ -148,7 +148,7 @@ const mixed = type => (el, next) => ({
  * @see BLOCK
  */
 const BLOCK_TAGS = {
-    caption: text('figure_caption'),
+    caption: text('caption'),
```
aiwenar: I'd prefer the default `<caption>` deserializer to check whether it's a child of `<figure>`, and skip it (return `undefined`) otherwise. This way we won't rely on a non-existent block (`'caption'`) and it should be easier to introduce a custom deserializer.

PiotrKozlowski:
> I'm not sure we want tables in admonitions or exercises, but this PR certainly shouldn't change default schema to include tables if they are being implemented outside `cnx-designer`.

In Introductory Statistics there is 89 tables inside notes and 214 tables inside exercises, so schema should be changed to accept those elements.
I don't see reason to create separate PR for updating schema for those elements or to do this from outside because it has to be here anyway.

src/cnxml/deserialize.js
PiotrKozlowski: When I'm doing this on the front:
```
function caption(el: Element, next: (nodes: any) => any) {
  if (el.parentElement && el.parentElement.tagName === 'TABLE') {
    return {
      type: 'table_caption',
      nodes: next(el.childNodes),
    }
  }
  return undefined
}
```
I'm getting `Error: `Block.fromJSON` requires a `type` string.`

I've tried with `return null`, `return undefined` and just `return`.

src/cnxml/deserialize.js
```diff
@@ -182,6 +182,21 @@ const MARK_TAGS = {
...
 
 
+/**
+ * Process data for captions.
+ */
+function caption(el, next) {
+    let type = 'caption'
+    if (el.parentElement) {
+        type = el.parentElement.tagName.toLowerCase() + '_caption'
+    }
```
aiwenar: Don't do that. You have no guarantee that CNXML element names will match Slate blocks one-to-one (e.g. `<note>` vs `admonition`), and no guarantee that future elements with captions will also want to deserialize them this way. Just ignore `<captions>` which are not inside a `<figure>`.

PR #21, opened 2019-04-09
Terms

PiotrKozlowski:
I'll handle definitions and glossary at separate branch so it's easier to review.

https://github.com/katalysteducation/adaptarr-front/pull/34

### DIFF
--- a/src/plugins/term/queries.js
+++ b/src/plugins/term/queries.js
@@ -27,11 +27,13 @@ export function getActiveTerm(change) {
   if (!focusLeaf.leaf.marks) return null
   focusLeaf.leaf.marks.some(m => {
     if (m && m.type === 'term') {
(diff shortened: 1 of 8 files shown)


src/plugins/term/queries.js
```diff
@@ -0,0 +1,45 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for
+// full license text.
+
+/* type Term = {
+  reference: string,
+  leaf: Slate~Leaf,
+  focusText: Slate~Text,
+  offsetStart: number,
+  offsetEnd: number,
+} */
```
aiwenar: You should define such types using JSDoc's virtual comments with [`@typedef`](http://usejsdoc.org/tags-typedef.html).

aiwenar:
Solved as inlines in #24

PR #26, opened 2019-04-25
Source edit

PiotrKozlowski:
Source edit for selected elements ('math', 'rule', 'statement', 'proof', 'equation', 'foreign')

https://github.com/katalysteducation/adaptarr-front/pull/38

### DIFF
--- a/src/cnxml/deserialize.js
+++ b/src/cnxml/deserialize.js
@@ -184,7 +184,6 @@ const mixed = type => (el, next) => ({
 const BLOCK_TAGS = {
     caption: caption,
     commentary: mixed('exercise_commentary'),
-    div: mixed('div'),
     exercise: 'exercise',
     figure: 'figure',
     image: image,
--- a/src/cnxml/deserialize.js
+++ b/src/cnxml/deserialize.js
@@ -83,15 +83,7 @@ const MARK = {
  */
 const DEFAULT = {
     deserialize(el, next) {
-        // We do not support fully those elements yet so for now they are 
-        // transformed into normal text so magicians can edit them in 
-        // source mode. 
-
-        const handleInSourceMode = [
-            'math', 'rule', 'statement', 'proof', 'equation', 'foreign'
-        ]
-
-        if (el.nodeName !== '#text' && handleInSourceMode.includes(el.localName.toLowerCase())) {
+        if (el.nodeName !== '#text' && SOURCE_TAGS.includes(el.localName.toLowerCase())) {
             const parentsForInlines = ['para', 'caption']
             let data = {
                 object: 'block',
@@ -138,6 +130,21 @@ const INLINE_TAGS = [
     'link',
     'sub',
     'sup',
+]
+
+/**
+ * We do not support fully those elements yet so for now they are 
+ * transformed into normal text so magicians can edit them in
+ * source mode.
+ */
+
+const SOURCE_TAGS = [
+    'equation',
+    'foreign',
+    'math',
+    'proof',
+    'statement',
+    'rule', 
 ]
 
 
--- a/src/cnxml/deserialize.js
+++ b/src/cnxml/deserialize.js
@@ -3,6 +3,7 @@
 // full license text.
 
 import { List } from 'immutable'
+import { Text } from 'slate'
 
 /**
  * Load a block element.
@@ -21,6 +22,8 @@ const BLOCK = {
             },
             nodes: next(Array.from(el.children)),
         }
+
+        if (props == null) return
 
         if (props instanceof Array) {
             props[0].key = props[0].key || el.getAttribute('id') || undefined
@@ -57,6 +60,8 @@ const MARK = {
             type: inline,
         }
 
+        if (props == null) return
+
         return {
             object: 'mark',
             key: el.getAttribute('id') || undefined,
@@ -78,6 +83,27 @@ const MARK = {
  */
 const DEFAULT = {
     deserialize(el, next) {
+        // We do not support fully those elements yet so for now they are 
+        // transformed into normal text so magicians can edit them in 
+        // source mode. 
+
+        const handleInSourceMode = [
+            'math', 'rule', 'statement', 'proof', 'equation', 'foreign'
+        ]
+
+        if (el.nodeName !== '#text' && handleInSourceMode.includes(el.localName.toLowerCase())) {
+            const parentsForInlines = ['para', 'caption']
+            let data = {
+                object: 'block',
+                type: 'source_element',
+                nodes: [Text.create(el.outerHTML)],
+            }
+            if (parentsForInlines.includes(el.parentElement.tagName)) {
+                data.object = 'inline'
+            }
+            return data
+        }
+
         if (el.nodeType !== Node.ELEMENT_NODE) {
             return
         }
@@ -158,6 +184,7 @@ const mixed = type => (el, next) => ({
 const BLOCK_TAGS = {
     caption: caption,
     commentary: mixed('exercise_commentary'),
+    div: mixed('div'),
     exercise: 'exercise',
     figure: 'figure',
     image: image,
--- a/src/cnxml/deserialize.js
+++ b/src/cnxml/deserialize.js
@@ -83,7 +83,7 @@ const MARK = {
  */
 const DEFAULT = {
     deserialize(el, next) {
-        if (el.nodeName !== '#text' && SOURCE_TAGS.includes(el.localName.toLowerCase())) {
+        if (el.nodeType !== Node.TEXT_NODE && SOURCE_TAGS.includes(el.localName.toLowerCase())) {
             const parentsForInlines = ['para', 'caption']
             let data = {
                 object: 'block',
--- a/src/cnxml/deserialize.js
+++ b/src/cnxml/deserialize.js
@@ -88,7 +88,7 @@ const DEFAULT = {
             let data = {
                 object: 'block',
                 type: 'source_element',
-                nodes: [Text.create(el.outerHTML)],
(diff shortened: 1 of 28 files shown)


PiotrKozlowski:
This should also wait for https://github.com/katalysteducation/slate-counters/pull/3

src/cnxml/deserialize.js
```diff
@@ -158,6 +184,7 @@ const mixed = type => (el, next) => ({
 const BLOCK_TAGS = {
     caption: caption,
     commentary: mixed('exercise_commentary'),
+    div: mixed('div'),
```
aiwenar: What do we know about `div`s?

src/cnxml/xml.js
```diff
@@ -61,6 +61,10 @@ class Renderer {
                 return Array.from(component, c => this.renderComponent(c))
             }
 
+            if (component instanceof Node) {
+                return component
```
aiwenar: `return this.doc.adoptNode(component)`

src/cnxml/serialize.js
```diff
@@ -142,6 +143,30 @@ function list(obj, children) {
...
 
 
+/**
+ * Serializer for source elements.
+ */
+function source(obj, children) {
+    const parser = new DOMParser();
+    const xmlDoc = parser.parseFromString(obj.getText().trim(), 'text/xml')
+    let el = document.createElement('template')
+    el.innerHTML = xmlDoc.documentElement.outerHTML
```
aiwenar: What you are doing here is, in order:

1. Deserialize XML (with `DOMParser`)

2. Serialize it as HTML (with `.outerHTML`)

3. Deserialize it as HTML (with `.innerHTML =`)

4. Use this HTML fragment as XML

why?

src/cnxml/serialize.js
```diff
@@ -142,6 +143,30 @@ function list(obj, children) {
...
+
+    const error = xmlDoc.getElementsByTagName('parsererror')
+    if (error.length) {
+        const attrs = {
+            class: 'invalid-xml',
+        }
+        return <div {...attrs}>
+            Error: {error[0].textContent}
+            Content: {obj.getText()}
+        </div>
```
aiwenar: If there are any errors with the XML fragment we want them reported immediately, not quietly. Documents with XML errors should not be allowed to save.

src/cnxml/serialize.js
```diff
@@ -142,6 +143,30 @@ function list(obj, children) {
...
+    }
+    
+    return <>{el.content.childNodes}</>
```
aiwenar: Why are you returning this as a React fragment, rather than just the element?

Also, why are you returning an array of elements, when there could only ever be one?

src/cnxml/xml.js
```diff
@@ -71,6 +75,9 @@ class Renderer {
 
     renderElement(element) {
         const ns = element.props.xmlns || this.ns
+        if (element.type.toString() == 'Symbol(react.fragment)') {
+            return Array.from(element.props.children).map(c => this.renderComponent(c))
+        }
```
aiwenar: `renderElement` accepts only React elements. If you want to add support for fragments then `renderComponent` is the place to do it.

Never case on a debug representation. While in this case it is stable (mandated by the ECMAScript standard), this check is nonetheless incorrect: while the correct symbol for fragments is (probably) `Symbol.for('react.fragment')` this check will also accept every `Symbol('react.fragment')`, which are different symbols. Note ‘are’: Symbol's constructor returns a new symbol every time it's called, regardless of arguments provided. To put it in code:  is false.

src/cnxml/deserialize.js
aiwenar: You've added support for deserializing `div`s, but havn't included it in schema and don't serialize it.

src/plugins/source/index.js
```diff
@@ -0,0 +1,11 @@
+// Copyright 2019 OpenStax Poland
+// Licensed under the MIT license. See LICENSE file in the project root for
+// full license text.
+
+import schema from './schema'
+import renderNode from './render'
+import onKeyDown from './handlers'
+
+export default function Source() {
```
aiwenar: This is not a good name.

src/plugins/source/render.js
```diff
@@ -0,0 +1,15 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for
+// full license text.
+
+import React from 'react'
+
+export default function renderNode({ node, children, attributes }, editor, next) {
+    if (node.type === 'source_element') {
+        return <span className={`source source--${node.object}`} {...attributes}>
+            {children}
+        </span>
```
aiwenar: Block nodes should not be rendered as `span`s.

test/cnxml/de/source-elements.js
```diff
@@ -0,0 +1,117 @@
...
+        </apply>
+    </math>
+</equation>
+`
+
+export const output = <value>
+    <document>
+        <source>
+{
+`<math xmlns="http://cnx.rice.edu/cnxml">
```
aiwenar: The correct namespace for MathML is `http://www.w3.org/1998/Math/MathML`

test/cnxml/de/source-elements.js
```diff
@@ -0,0 +1,117 @@
...
+    </proof>
+    <example id="pythag-example">
+        <para id="pythag-para3">
+        Take a right triangle whose sides are of length 
+        3, 4, and 5. In this case the sum of the square 
+        of the two shorter sides is 9+16=25. The square 
+        of the hypotenuse is 25. So the theorem holds.
+        </para>
+    </example>
+</rule>`
```
aiwenar: Do we really need to include complete fragments of books in tests? The larger the input data the bigger the diffs will be.

test/plugins/source/handle-backspace.js
```diff
@@ -0,0 +1,27 @@
...
+  editor.run('onKeyDown', { key: 'Backspace' })
+  editor.run('onKeyDown', { key: 'Backspace' })
+}
+
+export const input = <value>
+    <document>
+        <p>Before</p>
+        <source>
+            S<cursor/>ome
+            {'\n'}source
```
aiwenar: I don't think you should be able to move content between raw xml and normal document.

src/cnxml/deserialize.js
```diff
@@ -78,6 +83,27 @@ const MARK = {
...
+        const handleInSourceMode = [
+            'math', 'rule', 'statement', 'proof', 'equation', 'foreign'
+        ]
+
+        if (el.nodeName !== '#text' && handleInSourceMode.includes(el.localName.toLowerCase())) {
+            const parentsForInlines = ['para', 'caption']
+            let data = {
+                object: 'block',
+                type: 'source_element',
+                nodes: [Text.create(el.outerHTML)],
```
aiwenar: `el.outerHTML` will serialize the element as HTML, which is different from (and incompatible with) XML.

src/cnxml/serialize.js
PiotrKozlowski: How you would do this? I've tried few other options and that was the one that I found working.

test/cnxml/de/source-elements.js
PiotrKozlowski: It's working in editor but not in the tests. Do you know why it might be?

test/plugins/source/handle-backspace.js
PiotrKozlowski: I would like to keep consistency between code block and source elements as much as it is possible so it's easy for users to understand behaviour.

src/cnxml/deserialize.js
PiotrKozlowski: From what I've tested it's giving exactly the same output of element which was in the module content.

src/cnxml/serialize.js
aiwenar:
```
return new DOMParser().parseFromSting(obj.getText().trim(), 'application/xml`).documentElement
```

test/cnxml/de/source-elements.js
aiwenar: Because in test you create the input XML with wrong namespace.

src/cnxml/deserialize.js
aiwenar: One example: HTML has self-closing tags, XML doesn't. Using `<br>` as an example:

```js
root = document.createElement('root')
br = document.createElement('br')
root.appendChild(br)
new XMLSerializer().serializeToString(root)
// => "<root xmlns=\"http://www.w3.org/1999/xhtml\"><br /></root>"
root.outerHTML
// => "<root><br></root>"
new DOMParser().parseFromString(root.outerHTML, 'application/xml')
// => XML Parsing Error: mismatched tag. Expected: </br>.
```

PiotrKozlowski: But we are deserializing already parsed, valid xml?

aiwenar: `el` is not XML, it's a DOM element deserialized from XML. DOM has no knowledge of whether it came from HTML or XML, and it doesn't matter. `el.outerHTML` is a computed property which, whenever accessed, will serialize this particular DOM element to HTML or XML, depending on the type of document it came from (had to check reference to verify that, I was convinced it always serialized to HTML). Nonetheless it will always be safer to explicitly serialize to XML.

src/cnxml/serialize.js
aiwenar: > Why are you returning this as a React fragment, rather than just the element?

This is done because `slate-html-serializer` runs returned elements through React's `cloneElement` function, which corrupts non-react elements.

PiotrKozlowski:
Whole code for this plugin was moved into front end: https://github.com/katalysteducation/adaptarr-front/pull/43

PR #29, opened 2019-05-06
Handle enter inside figure_caption

PiotrKozlowski:
Handle enter inside figure_caption

### DIFF
--- a/src/plugins/figure/schema.js
+++ b/src/plugins/figure/schema.js
@@ -28,8 +28,10 @@ function normalizeFigure(change, error) {
     // and unwrap it
     case 'child_max_invalid':
         if (child.type === 'figure_caption') {
-            change.setNodeByKey(child.key, 'paragraph')
-            change.unwrapNodeByKey(child.key)
+            change.withoutNormalizing(() => {
+                change.setNodeByKey(child.key, 'paragraph')
+                change.unwrapNodeByKey(child.key)
+            })
         }
         break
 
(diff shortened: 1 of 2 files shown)


src/plugins/figure/schema.js
```diff
@@ -24,6 +24,15 @@ function normalizeFigure(change, error) {
         change.unwrapNodeByKey(child.key)
         break
 
+    // If there is more than 1 figure_caption then change second one to paragraph
+    // and unwrap it
+    case 'child_max_invalid':
+        if (child.type === 'figure_caption') {
+            change.setNodeByKey(child.key, 'paragraph')
+            change.unwrapNodeByKey(child.key)
```
aiwenar: When doing more than one operation, you need to wrap them in `change.withoutNormalizing`, otherwise slate wil renormalise the document between every operation. The only reason this works correctly is that any of those operations will work correctly on it's own: 

- just changing type to `paragraph` will trigger `child_unknown`, which will unwrap it;

- just unwrapping will create will trigger `parent_type_invalid` in `normalizeCaption`, which changes type to `paragraph`.

And the second operation (unwrapping) just happens to work:

- when figure is at document-level, caption cannot be unwrapped (there's nothing to unwrap it from), so it does nothing;

- when figure is in a section, `unwrapNodeByKey` will split section in two and move caption between these new sections, which triggers `next_sibling_type_invalid` in `normalizeSection` (since non-section after section is not valid), which moves it back into the section and merges the following  section back in.

You can see it for yourself by removing any of those two operations and running the test I've added.

PiotrKozlowski: I knew this will work without unwraping, but I didn't want to normalize code after first operation.
Thank for mentioning `withoutNormalizing` function.

PR #32, opened 2019-05-13
Fix for quotations

PiotrKozlowski:
- Insert paragraph when there is only `title` inside `quotation`.
- Change second `title` to `paragraph` - ex. when user hit `enter` inside `title`.
- Add test for first change and update test for duplicate titles.

### DIFF
--- a/src/plugins/quotation/schema.js
+++ b/src/plugins/quotation/schema.js
@@ -2,21 +2,40 @@
 // Licensed under the MIT license. See LICENSE file in the project root for
 // full license text.
 
+import { Text } from 'slate'
+
 function normalizeQuotation(change, error) {
     const { code, node, child, index } = error
 
     switch (code) {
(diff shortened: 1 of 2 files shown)


.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+  "editor.rulers": [80,120]
+}
```
aiwenar: Remove this file.

src/plugins/quotation/schema.js
```diff
@@ -2,21 +2,36 @@
...
+            case 'title':
+                change.insertNodeByKey(node.key, 1, {
+                    object: 'block',
+                    type: 'paragraph',
+                    nodes: [Text.create('')],
+                })
+                break
+            }
+            
             return
```
aiwenar: This will ignore errors not matching these criteria.

PR #33, opened 2019-05-13
Proper way of unwraping section

PiotrKozlowski:
`change.unwrapBlockByKey()` is not working in that case, so I'm unwraping content of section manually.

I've also added test which wasn't working with unwrapBlockByKey.


aiwenar:
This seems to me (shallow unwrapping) like a functionality that should be part of Slate core. You should probably bring it up on Slate's slack first.

PiotrKozlowski:
Okay, I've asked them.

PR #34, opened 2019-05-14
Fix captions

PiotrKozlowski:
- Section and Exercise can be inserted only into section or document.
- Invalid nodes inserted into `figure_caption` should be unwrapped. Paragraphs content should be kept as content of `figure_caption`

https://trello.com/c/LBuGZP0S/424-figure-caption-disappear

### DIFF
--- a/src/plugins/admonition/schema.js
+++ b/src/plugins/admonition/schema.js
@@ -34,6 +34,14 @@ function normalizeAdmonition(change, error) {
         console.warn('Unhandled admonition violation:', code)
         break
 
+    case 'child_type_invalid':
+        if (child.type === 'figure_caption') {
+            change.unwrapBlockByKey(node.key)
+            return
+        }
+        change.unwrapNodeByKey(child.key)
+        break
+
     default:
         console.warn('Unhandled admonition violation:', code)
         break
--- a/src/plugins/exercise/commands.js
+++ b/src/plugins/exercise/commands.js
@@ -24,6 +24,9 @@ export function insertExercise(change) {
             return !!document.getClosest(last.key, p2 => p1 === p2)
         }) || document
 
+        // Do not insert section in elements like exercise, admonition, etc.
+        if (parent.object !== 'document' && parent.type !== 'section') return
+
         // When using `change.wrapBlock` slate will place blocks according to its
         // own rules, which don't take into account legal parent-child relations,
         // which may cause exercise to be created as a child of e.g. a list,
--- a/src/plugins/figure/schema.js
+++ b/src/plugins/figure/schema.js
@@ -35,6 +35,14 @@ function normalizeFigure(change, error) {
         }
         break
 
+    case 'child_type_invalid':
+        if (child.type === 'paragraph') {
+            change.setNodeByKey(child.key, 'figure_caption')
+            break
+        }
+        change.unwrapBlockByKey(child.key)
+        break
+
     default:
         console.warn("Unhandled figure violation", violation)
         break
--- a/src/plugins/section/commands.js
+++ b/src/plugins/section/commands.js
@@ -19,6 +19,9 @@ export function insertSection(change) {
         const parent = document.getParent(start.key)
         const index = parent.nodes.indexOf(start)
 
+        // Do not insert section in elements like exercise, admonition, etc.
+        if (parent.object !== 'document' && parent.type !== 'section') return
+
         // Create title for the new section, and put cursor in it
         const title = Block.create({
             type: 'title',
--- a/src/plugins/section/schema.js
+++ b/src/plugins/section/schema.js
@@ -75,6 +75,12 @@ function normalizeSection(change, error) {
         console.warn('Unhandled section violation:', code)
         break
 
+    // Section was inserted into an invalid parent.
+    case 'parent_object_invalid':
+    case 'parent_type_invalid':
+        change.unwrapBlockByKey(node.key)
+        break
+
     default:
         console.warn('Unhandled section violation:', code)
         break
@@ -121,6 +127,7 @@ export default {
             ],
             next: { type: 'section' },
             normalize: normalizeSection,
+            parent: [{ object: 'document' }, { type: 'section' }],
         }
     },
 }
--- a/test/plugins/figure/insert-elements-in-caption.js
+++ b/test/plugins/figure/insert-elements-in-caption.js
@@ -1,0 +1,31 @@
+/** @jsx h */
+
+export default (change, editor) => {
+  change.insertAdmonition('note')
+  change.insertExercise()
+  change.wrapBlock('quotation')
+  change.insertSection()
+  change.wrapInList('ul_list')
+}
+
+export const input = <value>
+    <document>
+        <figure>
+            <media alt="First picture">
+                <img src="first.png"><text/></img>
+            </media>
+            <figcaption>Cap<cursor/>tion</figcaption>
+        </figure>
+    </document>
+</value>
+
+export const output = <value>
+    <document>
+        <figure>
+            <media alt="First picture">
+                <img src="first.png"><text/></img>
+            </media>
+            <figcaption>Cap<cursor/>tion</figcaption>
+        </figure>
+    </document>
+</value>


src/plugins/exercise/commands.js
```diff
@@ -24,6 +24,9 @@ export function insertExercise(change) {
             return !!document.getClosest(last.key, p2 => p1 === p2)
         }) || document
 
+        // Do not insert section in elements like exercise, admonition, etc.
+        if (parent.object !== 'document' && parent.type !== 'section') return
+
```
aiwenar: This check should be done in the UI, not here.

src/plugins/admonition/schema.js
```diff
@@ -34,6 +34,14 @@ function normalizeAdmonition(change, error) {
         console.warn('Unhandled admonition violation:', code)
         break
 
+    case 'child_type_invalid':
+        if (child.type === 'figure_caption') {
+            change.unwrapBlockByKey(node.key)
+            return
+        }
```
aiwenar: I don't understand what's the goal here. Why would you want to unwrap the entire note when it has a `figure_caption` child, but not when it has any other invalid child?

src/plugins/figure/schema.js
```diff
@@ -35,6 +35,14 @@ function normalizeFigure(change, error) {
         }
         break
 
+    case 'child_type_invalid':
+        if (child.type === 'paragraph') {
+            change.setNodeByKey(child.key, 'figure_caption')
+            break
+        }
+        change.unwrapBlockByKey(child.key)
+        break
```
aiwenar: `child_type_invalid` will only trigger if the invalid child is in between two valid children, otherwise `child_unknown` is triggered. In cases where it does trigger it will either split the figure in two, (if invalid child is not a paragraph) where one has no caption and the other is just caption, or (if invalid child is a paragraph) move the existing caption out of the figure (since changing paragraph to `figure_caption` will create two captions and trigger `child_max_invalid`, which will unwrap second (old) caption and convert it to a paragraph). Is this behaviour what you intended?

src/plugins/exercise/commands.js
PiotrKozlowski: Why not in both places considering that not anyone will use our front end?

src/plugins/admonition/schema.js
PiotrKozlowski: This is specific case when user insert note in figure_caption. Other invalid nodes are just unwrapped.

src/plugins/exercise/commands.js
aiwenar: Doing it only here will still allow users to press the button, and leave them wandering why nothing happened, so we need these checks in UI anyway. As for this function: commands are supposed to work even if this creates an invalid document, because it allows combining them into more complex changes within `withoutNormalizing`.

src/plugins/admonition/schema.js
aiwenar: This normalization (`child_type_invalid`) is triggered when `admonition` contains an invalid child, not when it is an invalid child. It will not trigger when an admonition is a child of `figure_caption`.

src/plugins/exercise/commands.js
PiotrKozlowski: > we need these checks in UI anyway

Yes, as we decided earlier.

> As for this function: commands are supposed to work even if this creates an invalid document, because it allows combining them into more complex changes within `withoutNormalizing`

This is something which we will probably never use.  Acctually those functions (inserting sections and exercises) are already very complex which is also probably not the best idea.

src/plugins/admonition/schema.js
PiotrKozlowski: This normalization is triggered when `figure_caption` ins wrapped with `note` after `change.insertAdmonition('note')`

src/plugins/figure/schema.js
PiotrKozlowski: In our editor this normalization is triggered only in specific situations like using `editor.wrapBlock('quotation')` inside `figure_caption` and then it's doing it's job. Other situations are unlikely to happen.

src/plugins/exercise/commands.js
aiwenar: > This is something which we will probably never use.

Maybe, maybe not, we don't know. And should we ever do, keeping things consistent will only benefit us.

> Acctually those functions (inserting sections and exercises) are already very complex which is also probably not the best idea.

They are nowhere near as complex as many Slate's build-in commands.

src/plugins/figure/schema.js
aiwenar: If you `wrapBlock('quotation')` a `figure_caption` then it would result in, I believe, a `figure` with two children: a `media` or `figure` followed by a `quotation` (which itself has a single child, a `figure_caption`). In such case `child_unknown` should be triggered.

src/plugins/exercise/commands.js
PiotrKozlowski: If we will start thinking of every possible case we will never finish version of Adaptarr which is usable by authors.

> They are nowhere near as complex as many Slate's build-in commands.

For me all `insert` functions should only insert elements. So `insertExercise` should just insert:
```
exercise
   problem
      para
```
Other functionalities are just complicating things. Like this with `figure_caption`, because we could handle this in normalization by unwraping invalid elements but `insertExercise` is trying to  handle all possible cases and is failing at this.

Solution is to not complicate things which doesn't have to be complicated.

src/plugins/admonition/schema.js
aiwenar: Oh, I see. I'd prefer we avoid making normalizations for such very specific cases. I agree that this particular normalization might be the best thing to do in such cases, but it arguably should never have happened in the first place; the UI should've prevented the user from making such change, and if it's a result of another legal change, then it's a bug. To sum up my approach to normalizations: they exist to fix documents which became invalid during a course of normal editing, not to best solve pathological cases which should never have happened.

src/plugins/exercise/commands.js
aiwenar: Normalizations exist to transform an invalid document into a valid one, not as helper utilities for transformations. They need to consider what could have caused an invalid document, and what's the simplest thing they can do to fix it. If you start using normalizations as helper utilities for transformations you'll quickly find situations where there are multiple possible “correct” normalizations and you don't know which to do as you lack context, or end up with an editor in which elements move around seemingly at random. Also, doing anything complex in normalizations will likely lead to code duplication, as you would have to re-create the same cases and transformations across many elements, possibly creating even more complex code due to sheer amount of possible child/parent combinations.

And even if you tried, it is impossible to move complexities of `insertExercise` into normalizations. It is written to support selection spanning multiple elements, and even ending at different depths.

aiwenar: As for calling `insertExercise` while selection is inside a `figure_caption`. In my opinion it works correctly: it removes `figure_caption` from figure, and wraps it in an exercise and a problem. The reason it doesn't work correctly is that `exercise_problem`'s `child_type_invalid` normalization is run before `figure_caption`'s `parent_type_invalid`.

src/plugins/figure/schema.js
PiotrKozlowski: Only `child_type_invalid` is called twice in this case.

PiotrKozlowski:
@aiwenar 
I'm a little bit confused how to approach this issue. You've picked options 1 and 3 for my question how should we solve this, which are:
> Nothing happens and additional validation on the front end to make sure no one is able to insert anything i captions.

If we want to have both of this then my understanding is to add specific rules in both: `cnx-designer` and `adaptarr-front`. Am I getting something wrong here?

src/plugins/exercise/commands.js
PiotrKozlowski: So why then this behaviour is correct for you and doing the same with note or quotation isn't?

aiwenar: You're right, we should extract this, let's call it pre-normalization, part from `insertExercise` into a separate command, and then use it also for inserting other complex structures.

aiwenar:
Sorry, I meant “nothing happens because there are checks in the UI”, and didn't realize 3 alone covers that.

PiotrKozlowski:
Okay, so then we need to block inserting functions depends on the current selection.

I don't have clear idea how to approach this.

We could get all parents types from current selection to the top of the document, ex. for selection inside figure caption it would be: `['figure_caption', 'figure', 'section']` and then pass this array to specific insert tool and compare it against given "permissions", like:
```
<Insert
   editor={editor}
   tool="admonition"
   parent={['figure_caption', 'figure', 'section']}
   disabledIn={['figure', 'table', 'quotation', ...]}
>
```

Do you think it could work? Would you approach this in other way?

aiwenar:
I think it would be simpler to pass just the list of legal parents, and check whether LCA of start and end of selection is in that list.

PiotrKozlowski:
In this case we are checking it for every insert tool, instead of checking it once and then only comparing array. Is this correct?

What LCA means?

aiwenar:
> In this case we are checking it for every insert tool, instead of checking it once and then only comparing array. Is this correct?

We can calculate LCA once and pass it as a prop, but otherwise I don't see any way not to check for every tool.

> What LCA means?

LCA stands for [Lowest Common Ancestor](https://en.wikipedia.org/wiki/Lowest_common_ancestor)

PiotrKozlowski:
LCA for this will be `para` which is valid for inserting exercise but is really invalid.
```
document
   note
      para
         <anchor/><focus/>
```
Should we omit `para` when calculating LCA?

But then what are valid parents for `note`? `['document', 'section']`?

aiwenar:
When inserting blocks I'd only consider block nodes. In this example `para` would be a leaf node, and so would be skipped.

Valid parents are: `document` and `section` for `admonition`, `exercise`, `example`, `definition`, etc.; `document`, `section, `exercise`, etc. for `*_list`, `quotation`, etc.

PiotrKozlowski:
Okay, I will prepare pr in `adaptarr-front` for this.

PiotrKozlowski:
Inserting blocks into invalid parent is now block by UI.

PR #35, opened 2019-05-17
Make TextContent and Document customizable

aiwenar:


src/plugins/index.js
```diff
@@ -50,24 +62,62 @@ export function TextContent(options={}) {
  *
  * This also includes {@link Content}
```
PiotrKozlowski: {@link TextContent}

src/plugins/index.js
```diff
@@ -50,24 +62,62 @@ export function TextContent(options={}) {
...
 export function Document(options={}) {
     const {
-        content = {},
         list = {},
     } = options
 
+    const content = [
+        ...(options.content || []),
+        'paragraph',
+        'quotation',
```
PiotrKozlowski: I'm not sure why `quotation` and `code` are here. `code` maybe make sens, but `quotation` is similar to `admonition`

aiwenar: Quotes were already allowed in all those places before this change. And personally I would consider it weird if adding quotes to exercises or notes was not allowed.

PR #36, opened 2019-05-21
Glossary v2

PiotrKozlowski:
- [x] Make `cnx-designer` works with `glossary`
- [x] Add `Definition` plugin
- [x] Adjust tests to work with `glossary`
- [x] Add tests for `definition`

UI: https://github.com/katalysteducation/adaptarr-front/pull/41

### DIFF
--- a/src/cnxml/glossary.js
+++ b/src/cnxml/glossary.js
@@ -1,0 +1,72 @@
+// Copyright 2019 OpenStax Poland
+// Licensed under the MIT license. See LICENSE file in the project root for
+// full license text.
+
+import React from 'react'
+
+/**
+ * De/serialization rules for document structure and elements.
+ */
+
+import { block, mixed } from './util'
+
+import { PARA } from './document'
+
+export const DEFINITION = block('definition', 'definition', 'definition', 'definition')
+
+function se_meaning(obj, children) {
+    let meaning_children = []
+    let examples = []
+
+    children.forEach(c => {
+        if (c.type === 'example') {
+            examples.push(c)
+        } else {
+            meaning_children.push(c)
+        }
+    })
+
+    if (examples.length === 0) {
+        return <meaning>
+            {meaning_children}
+        </meaning>
+    }
+
+    return <>
+        <meaning>
+            {meaning_children}
+        </meaning>
+        {examples}
+    </>
+}
+
+export const MEANING = block('meaning', mixed('definition_meaning'), 'definition_meaning', se_meaning)
+
+export const EXAMPLE = block('example', mixed('definition_example'), 'definition_example', 'example')
+
+export const SEEALSO = block('seealso', 'definition_seealso', 'definition_seealso', 'seealso')
+
+/**
+ * Process data for terms.
+ */
+function de_term(el, next) {
+    if (el.parentElement && el.parentElement.tagName === 'meaning') {
+        return
+    }
+
+    return {
+        type: 'definition_term',
+        nodes: next(el.childNodes),
+    }
+}
+
+export const DEFINITION_TERM = block('term', de_term, 'definition_term', 'term')
+
+export const GLOSSARY = [
+    DEFINITION,
+    MEANING,
+    EXAMPLE,
+    SEEALSO,
+    DEFINITION_TERM,
+    PARA,
+]
--- a/src/cnxml/index.js
+++ b/src/cnxml/index.js
@@ -9,6 +9,7 @@ import render from './xml'
 import { TEXT_CONTENT } from './text'
 import { DEFAULT } from './util'
 import { DOCUMENT } from './document'
+import { GLOSSARY } from './glossary'
 
 export function parseXml(xml) {
     const parsed = new DOMParser().parseFromString(xml, 'application/xml');
@@ -22,7 +23,15 @@ export function parseXml(xml) {
 }
 
 
-export function writeXml(content, options={}) {
+export function writeXml({document, glossary}, options={}) {
+    let glossaryContent = null
+
+    if (glossary) {
+        glossaryContent = <glossary>
+            {glossary}
+        </glossary>
+    }
+
     const r = <document
         xmlns="http://cnx.rice.edu/cnxml"
         cnxml-version="0.7"
@@ -32,8 +41,9 @@ export function writeXml(content, options={}) {
         >
         <title>TODO: load and preserve titles</title>
         <content>
-            {content}
+            {document}
         </content>
+        {glossaryContent}
     </document>
 
     return render(r, options)
@@ -41,9 +51,18 @@ export function writeXml(content, options={}) {
 
 
 export default class CNXML {
-    constructor(rules = []) {
+    constructor(args /* { documentRules: [], glossaryRules: [] } */) {
+        const {
+            documentRules = [],
+            glossaryRules = [],
+        } = args
         this.document = new Html({
-            rules: [...rules, ...DOCUMENT, ...TEXT_CONTENT, DEFAULT],
+            rules: [...documentRules, ...DOCUMENT, ...TEXT_CONTENT, DEFAULT],
+            defaultBlock: 'invalid',
+            parseHtml: x => x,
+        })
+        this.glossary = new Html({
+            rules: [...glossaryRules, ...GLOSSARY, ...TEXT_CONTENT, DEFAULT],
             defaultBlock: 'invalid',
             parseHtml: x => x,
         })
@@ -54,13 +73,17 @@ export default class CNXML {
             xml = parseXml(xml)
         }
 
-        return this.document.deserialize(find(xml.children, 'content'), options)
+        return {
+            document: this.document.deserialize(find(xml.children, 'content'), options),
+            glossary: this.glossary.deserialize(find(xml.children, 'glossary'), options),
+        }
     }
 
-    serialize(value, options={}) {
-        const content = this.document.serialize(value, { render: false })
+    serialize(documentValue, glossaryValue,  options={}) {
+        const document = this.document.serialize(documentValue, { render: false })
+        const glossary = glossaryValue ? this.glossary.serialize(glossaryValue, { render: false }) : null
 
-        return writeXml(content, options)
+        return writeXml({document, glossary}, options)
     }
 }
 
@@ -70,5 +93,8 @@ function find(children, name) {
         if (child.tagName === name) {
             return child
         }
+    }
+    return {
+        childNodes: [],
     }
 }
--- a/src/cnxml/xml.js
+++ b/src/cnxml/xml.js
@@ -50,6 +50,8 @@ class Renderer {
     }
 
     renderComponent(component) {
+        if (component == null) return null
+
         switch (component.$$typeof) {
         case REACT_ELEMENT_TYPE:
             if (component.type === React.Fragment) {
@@ -107,6 +109,7 @@ class Renderer {
     }
 
     appendChild(element, child) {
+        if (child == null) return
         if (child instanceof Node) {
             element.appendChild(child)
         } else if (child[Symbol.iterator]) {
--- a/src/cnxml/xml.js
+++ b/src/cnxml/xml.js
@@ -2,6 +2,8 @@
 // Licensed under the MIT license. See LICENSE file in the project root for
 // full license text.
 
+import React from 'react'
+
 /**
  * XML serializer, because `react-dom` enforces HTML constraints. For example,
  * React required that `<link>` tags have no children, contrary to what
@@ -50,6 +52,10 @@ class Renderer {
     renderComponent(component) {
         switch (component.$$typeof) {
         case REACT_ELEMENT_TYPE:
+            if (component.type === React.Fragment) {
+                return Array.from(component.props.children, c => this.renderComponent(c))
+            }
+
             return this.renderElement(component)
 
         default:
--- a/src/plugins/admonition/schema.js
+++ b/src/plugins/admonition/schema.js
@@ -34,6 +34,16 @@ function normalizeAdmonition(change, error) {
         console.warn('Unhandled admonition violation:', code)
         break
 
+    case 'child_type_invalid':
+        // Sometimes title is not first element. We want to change it to paragraph,
+        // because moving it to 0 index could be confusing for translators.
+        if (child.type === 'title') {
+            change.setNodeByKey(child.key, 'paragraph')
+            return
+        }
+        console.warn('Unhandled admonition violation:', code)
+        break
+
     default:
         console.warn('Unhandled admonition violation:', code)
         break
--- a/src/plugins/definition/handlers.js
+++ b/src/plugins/definition/handlers.js
@@ -1,0 +1,88 @@
+// Copyright 2019 OpenStax Poland
+// Licensed under the MIT license. See LICENSE file in the project root for
+// full license text.
+
+export default function onKeyDown(event, change, next) {
+    switch (event.key) {
+    case 'Enter':
+        return onEnter(event, change) || next()
+
+    case 'Backspace':
+        return onBackspace(event, change) || next()
+
+    default:
+        return next()
+    }
+}
+
+function onEnter(event, change) {
+    // Pressing shift should disable any special key handling
+    if (event.shiftKey) {
+        return false
+    }
+
+    const { value } = change
+    const { selection } = value
+    const block = value.startBlock
+
+    // In a definition.
+    const definition = change.getActiveDefinition(change.value)
+    if (!definition) return false
+
+    const item = definition.getParent(block.key)
+
+    if (block.type === 'definition_term' && item.type !== 'definition_seealso') {
+        change.moveToStartOfNextBlock()
+        return true
+    }
+
+    if (item.type === 'definition_example') {
+        // Only handle key if selection is in an empty block or at the beggining.
+        const isEmpty = selection.isCollapsed
+            && selection.start.isAtStartOfNode(block)
+            && selection.end.isAtEndOfNode(block)
+        if (!isEmpty && selection.start.offset > 0) return false
+
+        change.unwrapBlock('definition_example')
+        return true
+    }
+
+    if (item.type === 'definition_meaning') {
+        // Only handle key if selection is in an empty block or at the beggining.
+        const isEmpty = selection.isCollapsed
+            && selection.start.isAtStartOfNode(block)
+            && selection.end.isAtEndOfNode(block)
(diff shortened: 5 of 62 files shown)


src/plugins/definition/schema.js
```diff
@@ -0,0 +1,203 @@
...
+    }
+}
+
+function normalizeTerm(change, error) {
+    const { code, key, node, child } = error
+
+    switch (code) {
+    case 'parent_type_invalid':
+        change.unwrapBlockByKey(node.key)
+        break
```
aiwenar: This will insert raw text nodes into blocks which only accept blocks as children.

src/plugins/definition/schema.js
```diff
@@ -0,0 +1,203 @@
...
+        change.unwrapNodeByKey(child.key)
+        break
+
+    case 'child_unknown':
+        change.unwrapBlockByKey(child.key)
+        break
+
+    case 'parent_type_invalid':
+        change.unwrapBlockByKey(node.key)
+        break
```
aiwenar: This will move `node` one level up, where it will likely trigger the same normalization again, which will continue until it hits the document.

src/plugins/glossary/index.js
```diff
@@ -0,0 +1,15 @@
+// Copyright 2019 OpenStax Poland
+// Licensed under the MIT license. See LICENSE file in the project root for
+// full license text.
+
+import make_schema from './schema'
+
+export default function Glossary(options={}) {
+    const {
+        content = ['definition', 'paragraph'],
+    } = options
```
aiwenar: Glossary only contain definitions.

src/plugins/glossary/schema.js
```diff
@@ -0,0 +1,48 @@
...
+export default function schema(options) {
+    const content = options.content.map(type => ({ type }))
+
+    return {
+        document: {
+            nodes: [
+                {
+                    match: content,
+                    min: 1,
+                },
```
aiwenar: It should be OK for a glossary to contain no entries. There is a number of modules which have no glossaries.

src/plugins/section/schema.js
```diff
@@ -20,7 +20,11 @@ function normalizeDocument(change, error) {
         if (child.type === 'title') {
             change.setNodeByKey(child.key, { type: 'paragraph' })
             return
+        } else if (child.type === 'invalid') {
+            change.removeNodeByKey(child.key)
+            return
         }
```
aiwenar: What is this for?

test/util/cnxml.js
```diff
@@ -1,11 +1,14 @@
-export default global.cnxml = function cnxml(fragments) {
-    if (fragments.length > 1) {
-        throw new Error('cnxml`` does not support interpolation')
-    }
-
+export default global.cnxml = function cnxml(content, glossary) {
     return `<document xmlns="http://cnx.rice.edu/cnxml">
```
aiwenar: I don't like this change. It requires that we change all our serialization tests and write them in a more complex way, and the new addition is only ever used twice.

test/util/h.js
```diff
@@ -3,6 +3,11 @@ import { createHyperscript } from 'slate-hyperscript'
 export default global.h = createHyperscript({
     blocks: {
         code: 'code',
+        definition: 'definition',
+        determ: 'definition_term',
+        demeaning: 'definition_meaning',
+        deexample: 'definition_example',
+        deseealso: 'definition_seealso',
```
aiwenar: I know that we already have a two letter prefix (ex for exercise), but for some reason this is really confusing to me (possible because I've mostly seen de used as deserialization). I think I would prefer def as prefix here.

src/cnxml/index.js
```diff
@@ -32,18 +33,30 @@ export function writeXml(content, options={}) {
...
     return render(r, options)
 }
 
 
 export default class CNXML {
-    constructor(rules = []) {
+    constructor(args /* { documentRules: [], glossaryRules: [] } */) {
         this.document = new Html({
-            rules: [...rules, ...DOCUMENT, ...TEXT_CONTENT, DEFAULT],
+            rules: [...args.documentRules, ...DOCUMENT, ...TEXT_CONTENT, DEFAULT],
```
aiwenar: `args.documentRules` may be `undefined`.

src/cnxml/index.js
```diff
@@ -32,18 +33,30 @@ export function writeXml(content, options={}) {
...
-            {content}
+            {document}
         </content>
+        {
+            glossary ?
+                <glossary>
+                    {glossary}
+                </glossary>
+            : ''
+        }
```
aiwenar: I'd prefer this check outside of JSX. Normal ifs are much easier to read.

test/cnxml/index.js
```diff
@@ -29,32 +20,49 @@ global.DOMParser = dom.window.DOMParser
...
+    const resultContent = serialized.getElementsByTagName('content')[0]
+    const resultGlossary = serialized.getElementsByTagName('glossary')[0]
 
     const referenceXml = new DOMParser().parseFromString(output, 'application/xml')
-    const reference = referenceXml.querySelector(':root > content')
+    const referenceContent = referenceXml.querySelector(':root > content')
+    const referenceGlossary = referenceXml.querySelector(':root > glossary')
 
-    if (reference == null) {
+    if (referenceContent == null) {
```
aiwenar: This is a lot of additional complexity just for two tests.

test/plugins/glossary/definition/handle-backspace-in-meaning.js
```diff
@@ -0,0 +1,35 @@
...
+export const output = <value>
+    <document>
+        <definition>
+            <determ>
+                TermMeaning
+            </determ>
+            <demeaning>
+                <p>
+                    <cursor/>
+                </p>
```
aiwenar: Cursor should remain in front of “Meaning”.

test/plugins/glossary/definition/handle-backspace-in-term.js
```diff
@@ -0,0 +1,37 @@
+/** @jsx h */
+
+export default (change, editor) => {
+    editor.run('onKeyDown', { key: 'Backspace' })
+}
+
+export const input = <value>
+    <document>
+        <p>Text</p>
+        <definition>
```
aiwenar: Glossary can only contain definitions.

src/plugins/admonition/schema.js
```diff
@@ -34,6 +34,16 @@ function normalizeAdmonition(change, error) {
         console.warn('Unhandled admonition violation:', code)
         break
 
+    case 'child_type_invalid':
+        // Sometimes title is not first element. We want to change it to paragraph,
+        // because moving it to 0 index could be confusing for translators.
+        if (child.type === 'title') {
```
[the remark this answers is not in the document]
PiotrKozlowski: In module `History of Psychology` there is note with `para` at `0` index and then `title` at index `1`. When I moved it to `0` index then title makes no sens: https://cnx.org/contents/Sr8Ev5Og@10.16:lAYBvVZM@11/1-2-History-of-Psychology#fs-idm92314848

test/plugins/glossary/definition/handle-backspace-in-term.js
PiotrKozlowski: Hmm, ok. I will have to change how empty glossary behaves.

test/cnxml/index.js
PiotrKozlowski: What is your suggestion?

src/plugins/definition/handlers.js
```diff
@@ -0,0 +1,92 @@
...
+    const item = definition.getParent(block.key)
+
+    if (block.type === 'definition_term' && item.type !== 'definition_seealso') {
+        return change.moveToStartOfNextBlock()
+    }
+
+    if (item.type === 'definition_example') {
+        // Only handle key if selection is in an empty block
+        if (block.text.length) {
+            return
```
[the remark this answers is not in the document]
PiotrKozlowski: I probably can just check if `block.nodes.get(0)` have any leafs. This should be faster. Do you have other ideas?

src/plugins/definition/handlers.js
```diff
@@ -0,0 +1,92 @@
...
+    return
+}
+
+function onTab(event, change) {
+    const meaning = change.getActiveMeaning(change.value)
+    if (!meaning) return
+
+    event.preventDefault()
+
+    return change.insertExample()
```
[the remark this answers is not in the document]
PiotrKozlowski: Ah, yes. We were talking about this. I just think it's speeding up writing, but probably it should be other shortcut. I will delete this for now.

We probably should create bunch of shortcuts for this editor.

src/plugins/definition/schema.js
```diff
@@ -0,0 +1,203 @@
...
+    default:
+        console.warn('Unhandled definition_seealso violation:', code, JSON.stringify(child))
+        break
+    }
+}
+
+const CONTENT = [
+    { type: 'definition_meaning' },
+    { type: 'definition_example' },
+]
```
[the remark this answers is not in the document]
PiotrKozlowski: What is your suggestion here?

src/plugins/definition/schema.js
PiotrKozlowski: Meaning contains normal content like `paragraphs` so if there is invalid parent we are just unwraping meaning so content is not lost.

src/plugins/admonition/schema.js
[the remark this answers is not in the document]
aiwenar: This is actually not a paragraph but a `label` which gets converted into a paragraph, which will probably either be deleted and replaced with a class, or deleted and re-inserted (as `label`) before import to legacy.

test/util/cnxml.js
PiotrKozlowski: This was the simplest and fastest solution which came to my mind. You are welcome to propose other, however I don't see this being more complex, especially when I already adjusted tests and probably there will be not many more new functions coming.

src/plugins/definition/handlers.js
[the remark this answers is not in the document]
aiwenar: As far as I know there's currently no easy way to check whether a node is empty (there used to be, but it worked something like this code and was later removed for being a performance hazard (i.e. something users would use not knowing how much of a performance cost they are paying)). There is a place somewhere in exercise's commands or handlers where we check for this, so you should be able to just copy that code.

src/plugins/definition/schema.js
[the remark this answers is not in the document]
aiwenar: Something like this?
```js
blocks: {
    definition_meaning: {
        nodes: [
            { match: MEANING_CONTENT, min: 1 },
            { type: 'definition_example', min: 0 },
        ],
    },
},
```

src/plugins/definition/schema.js
aiwenar: It's the same mistake we did with sections. `unwrapBlockByKey` doesn't unwrap contents of the node, but the node itself from its parent.

test/cnxml/index.js
aiwenar: If you want to add more tests for glossary, set up a separate test function for them, otherwise just write those two tests without those helpers.

test/util/cnxml.js
aiwenar: The simplest solution would've been to just add a new function. As for it already being changed; this still makes tests harder to read and understand, and I don't agree that we won't change them in the future: Slate is still quire far from stable, and it's more than likely we'll have to adjust tests as it changes.

src/plugins/definition/schema.js
```diff
@@ -0,0 +1,203 @@
...
+    return {
+        blocks: {
+            definition: {
+                nodes: [
+                    { match: { type: 'definition_term' }, min: 1, max: 1 },
+                    { match: { type: 'definition_meaning' }, min: 1 },
+                    { match: CONTENT },
+                    { match: { type: 'definition_seealso' }, min: 0, max: 1 },
+                ],
+                first: { type: 'definition_term' },
```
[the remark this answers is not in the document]
PiotrKozlowski: But then error message is different: `child_min_invalid` vs current: `first_child_type_invalid` which is more descriptive.

src/plugins/section/schema.js
PiotrKozlowski: I guess I've added this while debugging. I'll delete this.

src/cnxml/index.js
PiotrKozlowski: You mean something like this:
```
let r
    
    if (glossary) {
        r = <document
            xmlns="http://cnx.rice.edu/cnxml"
            cnxml-version="0.7"
            id="new"
            module-id="new"
            xmlLang="en"
            >
            <title>TODO: load and preserve titles</title>
            <content>
                {document}
            </content>
            <glossary>
                {glossary}
            </glossary>
        </document>
    } else {
        r = <document
            xmlns="http://cnx.rice.edu/cnxml"
            cnxml-version="0.7"
            id="new"
            module-id="new"
            xmlLang="en"
            >
            <title>TODO: load and preserve titles</title>
            <content>
                {document}
            </content>
        </document>
    }
```

src/plugins/admonition/schema.js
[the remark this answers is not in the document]
PiotrKozlowski: Okay, but this doesn't change that moving this title to `0` index will be confusing for authors when they will work on adaptation. I think it's better to leave content as it is and let author decide if he want to add other title.

test/cnxml/index.js
PiotrKozlowski: I guess this will take even more time than what I've already did and it will not be simpler at all.

src/plugins/definition/handlers.js
```diff
@@ -0,0 +1,92 @@
...
+    const { value } = change
+    const { startBlock, selection } = value
+
+    const parent = definition.getParent(startBlock.key)
+
+    // Remove empty meanings
+    if (parent && parent.type === 'definition_meaning') {
+        const isEmpty = selection.isCollapsed
+        && selection.start.isAtStartOfNode(startBlock)
+        && selection.end.isAtStartOfNode(startBlock)
```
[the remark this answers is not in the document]
aiwenar: Disregard this comment, I misunderstood what this fragment was doing.

src/plugins/definition/handlers.js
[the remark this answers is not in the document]
PiotrKozlowski: I've used
```
const isEmpty = selection.isCollapsed
            && selection.start.isAtStartOfNode(block)
            && selection.end.isAtEndOfNode(block)
        if (!isEmpty && selection.start.offset > 0) return false
```

src/plugins/admonition/schema.js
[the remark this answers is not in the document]
aiwenar: OK. Lets leave it as it is then.

test/cnxml/index.js
```diff
@@ -43,24 +43,33 @@ function testDeserialization({ input, output }) {
...
-        const error = referenceXml.getElementsByTagName('parsererror')
-        throw new Error('Invalid XML:' + error[0].textContent)
+    if (contentValue) {
+        const resultContent = serialized.getElementsByTagName('content')[0]
+        const referenceContent = referenceXml.querySelector(':root > content')
+
+        if (referenceContent == null) {
+            const error = referenceXml.getElementsByTagName('parsererror')
+            throw new Error('Invalid XML:' + error[0].textContent)
+        }
```
aiwenar: This check should not be repeated in both if branches.

src/cnxml/glossary.js
```diff
@@ -12,7 +14,35 @@ import { PARA } from './document'
...
+        </meaning>
+    }
+
+    return <>
+        <meaning>
+            {meaning_children}
+        </meaning>
+        <example>
+            {example.props.children}
+        </example>
```
aiwenar: Meanings can have more than one example.

src/cnxml/xml.js
```diff
@@ -111,7 +113,9 @@ class Renderer {
             element.appendChild(child)
         } else if (child[Symbol.iterator]) {
             for (const node of child) {
-                this.appendChild(element, node)
+                if (node != null) {
+                    this.appendChild(element, node)
+                }
```
aiwenar: It would be better to make this check for all cases, not just here.

PiotrKozlowski: You mean like this:
```
appendChild(element, child) {
        if (child == null) return
        if (child instanceof Node) {
            element.appendChild(child)
        } else if (child[Symbol.iterator]) {
            for (const node of child) {
                if (node != null) {
                    this.appendChild(element, node)
                }
            }
        } else {
            throw new Error("invalid child")
        }
    }
```
And maybe one more at the beggining, right there: ?
```
const root = this.renderComponent(tree)
```

aiwenar: > You mean like this:

Yes, but without the nested check (within `for (const node of child)`).

> And maybe one more at the beggining, right there: ?

You mean a check that the root node is not null? I think we can safely assume it's not.

src/cnxml/glossary.js
```diff
@@ -36,9 +36,13 @@ function se_meaning(obj, children) {
...
         </meaning>
-        <example>
-            {example.props.children}
-        </example>
+        {
+            examples.map(e => {
+                return <example>
+                    {e.props.children}
+                </example>
+            })
```
aiwenar: What is this map for? Shouldn't `examples` already be wrapped in `<example>`?

PiotrKozlowski: :man_facepalming: True

test/util/cnxml.js
PiotrKozlowski: I've rolled back those changes and adjusted tests, but I still consider this not intuitive since we allow two values in out main `CNXML` class.

src/plugins/definition/schema.js
```diff
@@ -0,0 +1,203 @@
...
+    switch (code) {
+    case 'child_min_invalid':
+        change.removeNodeByKey(node.key)
+        break
+    
+    case 'child_type_invalid':
+        if (Text.isText(child)) {
+            change.wrapNodeByKey(child.key, Block.create('paragraph'))
+            break
+        }
```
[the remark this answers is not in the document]
PiotrKozlowski: I've deleted those checks. They were written to make sure that invalid cnxml is transformed well, but we've agreed that normalizations should handle cases which might happen when working with document.

aiwenar:
@PiotrKozlowski In de/serialization tests please change `export const input/output` from an object into two exports and I think it'll be ready to clean and merge.

PR #39, opened 2019-06-03
Code coverage

aiwenar:
Add code coverage, additional tests, and fix some behaviours.

@PiotrKozlowski Could you add additional tests for uncovered normalizations in your plugins?

### DIFF
--- a/src/plugins/admonition/schema.js
+++ b/src/plugins/admonition/schema.js
@@ -2,8 +2,6 @@
 // Licensed under the MIT license. See LICENSE file in the project root for
 // full license text.
 
-import { List } from 'immutable'
-
 // Admonition types supported by both CNXML's <note> and HTMLBook. Missing are
 // `aside` from CNXML and `caution` from HTMLBook.
 const TYPES = ["note", "warning", "tip", "important"]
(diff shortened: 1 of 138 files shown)


src/plugins/admonition/schema.js
```diff
@@ -14,13 +12,6 @@ function normalizeAdmonition(change, error) {
     switch (code) {
     // Admonition type is not valid.
     case 'node_data_invalid':
-        if (key === 'class') {
```
PiotrKozlowski: Why should we omit this fix? I know that this will not happen in our application because of front end validation, but shouldn't `cnx-designer` be standalone package?

PiotrKozlowski: nvm, there is separate Plugin for this :D

aiwenar: I've extracted classes to a separate plugin as it limits code duplication and makes it easier to add support for new elements and to test it (since we now only need to test it once, and not once per element).

src/plugins/definition/schema.js
```diff
@@ -60,33 +60,19 @@ function normalizeDefinition(change, error) {
...
-
     /* istanbul ignore next */
     default:
         console.warn('Unhandled definition violation:', code)
         break
     }
 }
 
+/* istanbul ignore next */
 function normalizeTerm(change, error) {
```
aiwenar: Don't ignore functions, it will be very easy to forget about it when we add some normalizations.

src/plugins/code/schema.js
```diff
@@ -3,25 +3,29 @@
 // full license text.
 
 function normalizeCodeBlock(change, error) {
```
PiotrKozlowski: It seems like this function is never called and I can't figure out why.

PR #40, opened 2019-06-04
Slate upgrade to 0.47.0

aiwenar:


src/plugins/term/index.js
```diff
@@ -2,7 +2,7 @@
 // Licensed under the MIT license. See LICENSE file in the project root for
 // full license text.
 
-import renderNode from './render'
+import renderDecoration from './render'
```
PiotrKozlowski: renderInline

src/plugins/text/render.js
```diff
@@ -26,7 +26,7 @@ export function renderNode({ node, children, attributes }, editor, next) {
     return next()
 }
 
-export function renderMark({ mark, children, attributes }, editor, next) {
+export function renderDecoration({ mark, children, attributes }, editor, next) {
```
PiotrKozlowski: It's rendering `span` instead correct mark tag

PiotrKozlowski: You can still use `renderMark` here. From what I understand it's no longer used for rendering decoration, but it's still correct function.

PR #43, opened 2019-06-10
Keep node.key for insert_node operation in IndexDB

PiotrKozlowski:
https://trello.com/c/K6AiJ3ZB/465-when-user-refresh-draft-elements-keys-are-chaning

<details>
<summary>Click t show trello card content</summary>
`Insert exercise`
`Insert reference` - to this exercise
Reference is correct
`Refresh page`
Reference is undefined, because key of new inserted exercise has changed.

Local changes in IndexDB for inserting nodes should also save element's key and restore it properly.
</details>

### DIFF
--- a/src/persistence/index.js
+++ b/src/persistence/index.js
@@ -189,9 +189,15 @@ export class DocumentDB {
         const tx = this.database.transaction('changes', 'readwrite')
         const store = tx.objectStore('changes')
 
+        let operation = op.toJS()
+        if (op.type === 'insert_node' && op.node.object === 'block') {
+            // We need to keep the same key for inserted nodes so xrefs targets
+            // are still valid after refreshing page.
+            operation.node.key = op.node.key
+        }
         await promisify(store.add({
             document: this.id,
-            change: op.toJS(),
+            change: operation,
         }))
     }
 


src/persistence/index.js
```diff
@@ -189,9 +189,15 @@ export class DocumentDB {
         const tx = this.database.transaction('changes', 'readwrite')
         const store = tx.objectStore('changes')
 
+        let operation = op.toJS()
+        if (op.type === 'insert_node' && op.node.object === 'block') {
+            // We need to keep the same key for inserted nodes so xrefs targets
+            // are still valid after refreshing page.
+            operation.node.key = op.node.key
+        }
```
aiwenar: `op.node` is not actually a node, it's a set of properties that will be used to build the node when the operation is applied, and most importantly it does _not_ include key. The key will be generated when the operation is applied, which is why `slate-counters` was getting it from the next value (`nextOp.value.document.getNode(op.path).key`), but as of Slate 0.46 it just gives up as there is no way of retrieving it.

PiotrKozlowski: So acctually from mine tests it does include `key`, but when you do `op.node.toJS()` then `key` is not there.

This `key` is important for us since local counters will crush if it changes after refresh.

aiwenar: I've just checked that what I said is true for `split_node`, but it seem that `insert_node` does in fact respect `op.node.key` if set. However, if `Operation.toJS` doesn't return `key` then we should not expect that `Operation.apply` will respect it.

Also do note that Slate will remove keys in a few versions (ianstormtaylor/slate#2864).

PiotrKozlowski: Ouch.

I would merge this pr for now since it's helping us with current version of counters and do something with it when keys will be removed - this will be big change for us since we will have to figure out how to target references.

PR #49, opened 2019-07-02
Do not allow dragging figures

PiotrKozlowski:
Since dragging figures is causing weird things it's easier to just disable this option.

https://trello.com/c/iHmQ6u5K/310-disable-dragging-or-fix-figure-dragging


aiwenar:
You should never use JS strings in attributes.

PiotrKozlowski:
Feel free to change it to whatever you like or suggest what it should look like

PR #50, opened 2019-07-02
Fix subfigures

PiotrKozlowski:
When user remove one of subfigures, the second one should be unwrapped.
https://trello.com/c/Vjj2rbgO/302-subfigure-error

There is also new error with slate-counters, which is counting subfigures like this:
```
Figure 1
   Figure 2
   Figure 3
```
https://trello.com/c/wg0m9oFd/501-wrong-counters-for-subfigures

### DIFF
--- a/src/plugins/figure/commands.js
+++ b/src/plugins/figure/commands.js
@@ -56,34 +56,36 @@ export function insertSubfigure(change, media) {
         return
     }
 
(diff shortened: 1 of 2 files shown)


aiwenar:
All this has already been done in #3 (which we should probably just merge without waiting for Slate).

PiotrKozlowski:
It's your call. This will work with current version for sure

PR #51, opened 2019-07-02
Fix titles

PiotrKozlowski:
> When switching a paragraph into a title in a section this paragraph is removed from the document.

- I've add optional `titleText` param to `insertFunction` which will cause creating section with given text for title.
- I've changed schema to handle cases when title is not first node.

~~However I can't figure out why `undo` is not working.~~

https://trello.com/c/PAKHEIG5/502-title-error-in-section

### DIFF
--- a/src/plugins/section/commands.js
+++ b/src/plugins/section/commands.js
@@ -8,8 +8,9 @@ import { Block, Range, Text } from 'slate'
  * Create a new section.
  *
  * @param {Slate~Change} change
+ * @param {string?} titleText
  */
-export function insertSection(change) {
+export function insertSection(change, titleText) {
     change.withoutNormalizing(change => {
         const { value } = change
         const { document, startBlock, endBlock } = value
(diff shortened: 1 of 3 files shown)


src/plugins/section/commands.js
```diff
@@ -8,8 +8,9 @@ import { Block, Range, Text } from 'slate'
  * Create a new section.
  *
  * @param {Slate~Change} change
+ * @param {string} titleText
  */
-export function insertSection(change) {
+export function insertSection(change, titleText = "") {
```
aiwenar: I'd prefer the decision what to do if title text isn't specified was left to Slate.

src/plugins/section/schema.js
```diff
@@ -59,6 +59,25 @@ function normalizeSection(change, error) {
...
+        } else if (child.type === 'title') {
+            // This usually happens when title occure in middle of section after changing
+            // block type from paragraph to title.
+
+            // Create new section with given title and remove unused child.
+            const text = child.text
+            change.withoutNormalizing(() => {
+                change.insertSection(text)
+                change.removeNodeByKey(child.key)
+            })
```
aiwenar: This will discard any marks from the title.

I think it would be better to implement it as
```js
change.wrapBlockByKey(child.key, 'section')
const section = change.value.document.getParent(child.key)
change.unwrapNodeByKey(section.key)
```
(note that there is purposefully no `withoutNormalizing`; we want normalizations to run after `wrapBlockByKey` so that they move all nodes after title into the new section).

PR #52, opened 2019-07-04
Implement whitespace normalization

aiwenar:


PiotrKozlowski:
In `Applied Probability` book there is module `Minterms and MATLAB Calculations` in chapter `
Minterm Analysis`.

When I was opening before there was an error: `Error: 'Mark.fromJS' requires a 'type' string.`

Now there is an error: `Error: Cannot set on an immutable record.`

In `Preface to Pfeiffer Applied Probability` there is still `Mark.fromJS ...` error.

I'm not sure if those errors are related to whitespaces for sure, but I guess this might be related to: https://trello.com/c/saNhVlK1/411-bullet-points-error

PiotrKozlowski:
In `Preface to Pfeiffer Applied Probability` it's failing at text `Appendix B` which in `cnxml` is `<link document="m23990">Appendix B</link>` but slate is trying to serialize this as:
```
{
  "object": "text",
  "text": "Appendix B",
  "marks": [
    Object { type: undefined, data: undefined }
  ]
}
```
which is causing `'Mark.fromJS' requires a `type` string.'`

This error is probably not related to this pr.

src/cnxml/whitespace.js
```diff
@@ -0,0 +1,279 @@
...
+            // 5th step
+            .replace(/[\u0020\u1680\u2000-\u2006\u2008\u2009\u200A\u205F\u3000]+([\u00a0])/g, (_, r) => r)
+            .replace(/([\u00a0][\u0020\u1680\u2000-\u2006\u2008\u2009\u200A\u205F\u3000]+)/g, (_, r) => r)
+            // 6th step
+            .replace(/[\s\u180e\u200b\u200c\u200d\u2060]{2,}/g, c => c[0])
+    }
+
+    switch (node.object) {
+    case 'text':
+        node.text = replaceWS(node.text)
```
PiotrKozlowski: This is failing for:
```
Object { object: "text", text: "<m:math xmlns:m=\"http://www.w3.org/1998/Math/MathML\" overflow=\"scroll\"><m:mrow><m:mi>A</m:mi><m:mo>,</m:mo><m:mspace width=\"0.166667em\"/><m:mi>B</m:mi><m:mo>,</m:mo><m:mspace width=\"0.166667em\"/><m:mi>C</m:mi></m:mrow></m:math>", marks: [] }
```
`Error: Cannot set on an immutable record.`

PiotrKozlowski: This is because `node` is created as immutable object instead of normal js object: https://github.com/katalysteducation/adaptarr-front/blob/master/src/screens/app/Draft/plugins/SourceElements/deserialize.ts#L15

PR #53, opened 2019-07-05
Fix rollup

PiotrKozlowski:


aiwenar:
Fixed in 66c8241, since it didn't compile for me after removing `'@babel/runtime/helpers/objectSpread'` (it would seem we need both `objectSpread` and `objectSpread2`).

PR #56, opened 2019-07-18
Preserve title and language

PiotrKozlowski:
https://github.com/openstax-poland/adaptarr-front/pull/108

### DIFF
--- a/src/cnxml/index.js
+++ b/src/cnxml/index.js
@@ -24,6 +24,11 @@ export function parseXml(xml) {
 
 
 export function writeXml({document, glossary}, options={}) {
+    const {
+        title = '',
+        language = 'en',
+    } = options
+
     let glossaryContent = null
 
     if (glossary) {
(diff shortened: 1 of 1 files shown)


src/cnxml/index.js
```diff
@@ -74,6 +79,7 @@ export default class CNXML {
         }
 
         return {
+            language: xml.getAttribute('xml:lang'),
```
aiwenar: `xml.getAttributeNS('http://www.w3.org/XML/1998/namespace', 'lang')`

src/cnxml/index.js
```diff
@@ -24,6 +24,11 @@ export function parseXml(xml) {
 
 
 export function writeXml({document, glossary}, options={}) {
+    const {
+        title = 'No title provided',
+        language = 'en',
```
aiwenar: If title wasn't provided the default should be empty.

PR #57, opened 2019-07-19
Add toJS() and replaceDatabase() to PersistDB

PiotrKozlowski:
Working version of UI which is based on those changes: https://github.com/openstax-poland/adaptarr-front/pull/110

### DIFF
--- a/src/persistence/index.js
+++ b/src/persistence/index.js
@@ -76,6 +76,103 @@ export class PersistDB {
         return PersistDB.open().then(db => db.openDocument(id))
     }
 
+    /**
+     * Export contents of this database as a JS object.
+     *
+     * Returned object can later be used to import data into a database using
+     * {@link #import}. It contains only plain JS values and can safely be
+     * converted to/from JSON.
+     */
+    async export() {
+        const tx = this.database.transaction(
+            this.database.objectStoreNames, 'readonly')
+
+        const insert = {}
+        const objectStores = {}
+
+        for (const name of this.database.objectStoreNames) {
+            const store = tx.objectStore(name)
+            const indexes = Object.fromEntries(Array.from(store.indexNames)
+                .map(name => {
+                    const index = store.index(name)
+                    return [name, {
+                        name: index.name,
+                        keyPath: index.keyPath,
+                        multiEntry: index.multiEntry,
+                        unique: index.unique,
+                    }]
+                }))
+
+            insert[name] = await promisify(store.getAll())
+
+            objectStores[name] = {
+                indexes,
+                keyPath: store.keyPath,
+                autoIncrement: store.autoIncrement,
+            }
+        }
+
+        return {
+            database: {
+                name: this.database.name,
+                version: this.database.version,
+                objectStores,
+            },
+            remove: Object.fromEntries(
+                Array.from(this.database.objectStoreNames).map(st => [st, {}])),
+            insert,
+        }
+    }
+
+    /**
+     * Import data into this database.
+     */
+    async import(data) {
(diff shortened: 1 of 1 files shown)


src/persistence/index.js
```diff
@@ -76,6 +76,105 @@ export class PersistDB {
         return PersistDB.open().then(db => db.openDocument(id))
     }
 
+    /**
+     * Return JS Object of IndexedDB with DB_NAME.
+     */
+    static toJS() {
+        return PersistDB.open().then(async db => {
+            const tx = db.database.transaction(
+                ['states', 'changes', 'contents'], 'readonly')
```
aiwenar: I'd prefer this code to be independent from current implementation, for example using database properties such as `IDBDatabase.objectStoreNames`, `IDBDatabase.name`, or `IDBDatabase.version`.

src/persistence/index.js
```diff
@@ -76,6 +76,105 @@ export class PersistDB {
...
+     * ot with new content.
+     *
+     * If database does not exists, it will be created.
+     *
+     * If you provide @param ids then this will remove only data for those ids.
+     *
+     * @param {Object} newDB
+     * @param {string[]} ids
+     */
+    static replaceDatabase(newDB = {}, ids = []) {
```
aiwenar: This should be an asynchronous method.

src/persistence/index.js
```diff
@@ -76,6 +76,105 @@ export class PersistDB {
...
+            await Promise.all([newChanges.map(ch => {
+                return promisify(changes.put(ch))
+            })])
+
+            await Promise.all([newContents.map(ct => {
+                return promisify(contents.put(ct))
+            })])
+        }).catch(e => {
+            throw new Error(e)
+        })
```
aiwenar: Since this promise isn't returned, this will cause all errors to be silently ignored.

src/persistence/index.js
```diff
@@ -76,6 +76,105 @@ export class PersistDB {
...
+            const changes = tx.objectStore('changes')
+            const contents = tx.objectStore('contents')
+
+            if (ids.length) {
+                let promises = []
+                ids.map(id => {
+                    promises.push(promisify(states.delete(id)))
+                    promises.push(promisify(contents.delete(id)))
+                })
+                await Promise.all([
```
aiwenar: This could be simplified to `await Promise.all(ids.map(id => db.openDocument(id).then(d => d.discard())))`.

src/persistence/index.js
```diff
@@ -76,6 +76,105 @@ export class PersistDB {
...
+                    states.clear(),
+                    changes.clear(),
+                    contents.clear(),
+                ].map(promisify))
+            }
+
+            await Promise.all([newStates.map(st => {
+                return promisify(states.put(st))
+            })])
```
aiwenar: `Promise.all` accepts an array of promises, but here you pass it an one-element array containing an array of promises.

src/persistence/index.js
PiotrKozlowski: When I'm doing this then I'm getting this error:
`Error: "TransactionInactiveError: A request was placed against a transaction which is currently not active, or which is finished."` for:
```
await Promise.all(statesToUpdate.map(st => 
    promisify(states.put(st))
))
```

src/persistence/index.js
PiotrKozlowski: I'm not sure about this. It's easier to manage when we do exactly what we want and when we will update database we just have to remember to update also those changes, but I'm open to suggestions ho to make this universal.

aiwenar:
I've simplified and restyled code a bit, if it looks OK to you you can squash changes and it'll be ready to merge.

PR #58, opened 2019-07-24
Throw a custom error when restoration from DocumentDB fails

aiwenar:


src/persistence/index.js
```diff
@@ -239,6 +245,15 @@ export class DocumentDB {
     }
 }
 
+export class RestoreError extends Error {
+    constructor(cause, value) {
+        super(`Cannot restore: ${cause.message}`)
+
```
PiotrKozlowski: // Set the prototype explicitly
Object.setPrototypeOf(this, PersistanceError.prototype)

PR #59, opened 2019-08-09
Set this.dirty to false on DocumentDB.discard()

PiotrKozlowski:


aiwenar:
The flag should only be updated after changes have been discarded.

PR #60, opened 2019-09-03
Image texts

PiotrKozlowski:
This pr is moving alt text from attribute to the separate block so it's easier to write for authors and there is possibility for review and suggest changes.
There is also new block allowed - `media_text` which authors can use to translate text on image.

Both of those blocks are saved as `divs` in `cnxml`. Other solution would be to create separate namespace and custom tag names for them.

Holding those in separate blocks mean that magicians will have to remove `media_text` block when image is replaced with correct text and manually move `media_alt` text to the `alt` attribute.

`media_alt` can't be moved automatically to the `<media alt="">` because it can contain `suggestions`.
If you have idea how it can be solved differently, let me know.

Front: https://github.com/openstax-poland/adaptarr-front/pull/153


src/cnxml/document.js
```diff
@@ -156,17 +156,67 @@ export const LIST_ITEM = block('item', de_list_item, 'list_item', 'item')
...
+    }
+}
+
+/**
+ * Deserializer for divs.
+ */
+function de_div(el, next) {
+    const type = el.getAttribute('class') || 'div'
+
+    return mixedContent(el, next, type.replace('-', '_'))
```
aiwenar: We should only support loading of `div`s with known class and properties, otherwise you can use them to insert any content into the model.

src/plugins/media/schema.js
```diff
@@ -1,44 +1,43 @@
...
+            image: {
+                isVoid: true,
+                data: {
+                    src: Boolean,
+                },
+            },
+            media_alt: {
+                marks: marks.map(type => ({ type })),
+            },
+            media_text: {
```
aiwenar: Keys here should be alphabetically sorted.

src/plugins/media/schema.js
```diff
@@ -1,44 +1,43 @@
...
+            // TODO: if we do, add target attribute.
+            image: {
+                isVoid: true,
+                data: {
+                    src: Boolean,
+                },
+            },
+            media_alt: {
+                marks: marks.map(type => ({ type })),
+            },
```
aiwenar: Since alt-text is ultimately going to be stored in the `alt` attribute it can't contain marks.

test/util/compareHtml.js
```diff
@@ -113,6 +113,8 @@ const BLOCK_NODES = [
     'figure',
     'list',
     'media',
+    'media_alt',
+    'media_text',
```
aiwenar: `media_alt` and `media_text` are Slate nodes and not CNXML elements, and as such they will never occur in CNXML. `div` on the other hand will and should be included in `BLOCK_NODES`.

test/cnxml/se/figure.js
```diff
@@ -13,8 +14,9 @@ export const inputContent = <value>
 
 export const output = cnxml`
 <figure id="f1">
-    <media alt="Figure content">
+    <media>
         <image src="f1.png" />
+        <div class="media-alt">Figure content</div>
```
aiwenar: Serialization in this case should remain unchanged.

test/cnxml/se/figure-subfigures.js
```diff
@@ -22,14 +24,16 @@ export const inputContent = <value>
...
             <image src="f1.png" />
+            <div class="media-alt">First subfigure's content</div>
         </media>
         <caption>First subfigure</caption>
     </subfigure>
     <subfigure id="f1-2">
-        <media alt="Second subfigure's content">
+        <media>
             <image src="f2.png" />
+            <div class="media-alt">Second subfigure's content</div>
```
aiwenar: Serialization in these cases should remain unchanged.

src/plugins/media/render.js
```diff
@@ -16,6 +16,22 @@ export default function renderBlock({ node, attributes, isFocused }, editor, nex
             isFocused={isFocused}
             />
 
+    case 'media_alt':
+        return <div
+            className="media-alt"
+            attributes={attributes}
+            >
+                {children}
+            </div>
```
aiwenar: Style: closing tag should be aligned to statement, and content indented once. See for example https://github.com/openstax-poland/cnx-designer/blob/master/src/plugins/section/render.js#L10-L12.

src/plugins/media/schema.js
```diff
@@ -1,44 +1,43 @@
...
-            data: {
-                alt: Boolean,
+export default function schema({ marks }) {
+    return {
+        blocks: {
+            // TODO: do we actually want to keep multiple versions?
+            media: {
+                nodes: [
+                    { match: { type: 'image' }, min: 1 },
+                    { match: { type: 'media_alt' }, min: 0, max: 1 },
```
aiwenar: As with the old `alt` attribute, `media_alt` must always be present.

PiotrKozlowski:
> That's an internal detail of our current process, and as such I think I would prefer it to be implemented in the front-end.

Do you know if there is easy way to merge schemas for plugins?
If not then probably we should rewrite whole support for figures into front end?

> The alt attribute should always be generated. In case where alt text contains suggestions they should be stripped when generating the attribute (that is deletions should be cancelled, and additions ignored).

This means that you can't suggest changes in alt text, because you have to serialize document before changing step / saving document and from what I remember that was one of acceptance criteria.

aiwenar:
> Do you know if there is easy way to merge schemas for plugins?

Actually, Slate merges schemas by default. The schema format we use (`blocks: { type: { /* schema */ } }`) has been deprecated and is being kept for compatibility and as a shorthand, and is being automatically translated into the new format, which is an array of schema object with a `match` property describing to which nodes it applies (see https://github.com/openstax-poland/cnx-designer/blob/figure-tests/src/plugins/figure/schema.js#L115-L139 for an example). So you should be able to just do something like

```
{
    rules: [
        {
            match: { type: 'figure' },
            last: {
                object: 'block',
                type: 'media_text',
            },
        },
    ],
}
```

PiotrKozlowski:
Now there is support only for `media_alt` which is converted always to `alt` attribute.

If in `media_alt` were only `text` nodes then this block will not be saved in cnxml. If some `inlines` occured then new `div` with `media-alt` class will be added inside `media` element.

Since suggestions are not part of `cnx-designer` I think there is no good reason to try resolve them before creating alt text so I'm just using `mediaAlt.text` when creating attribute.

src/cnxml/document.js
```diff
@@ -156,17 +156,90 @@ export const LIST_ITEM = block('item', de_list_item, 'list_item', 'item')
...
+
+    if (!hasAltDiv) {
+        nodes.push({
+            object: 'block',
+            type: 'media_alt',
+            nodes: [{
+                object: 'text',
+                text: alt,
+            }],
+        })
```
aiwenar: This is already being done in schema normalizations.

src/cnxml/document.js
```diff
@@ -156,17 +156,90 @@ export const LIST_ITEM = block('item', de_list_item, 'list_item', 'item')
...
 }
 
-export const MEDIA = block('media', media, 'media', 'media')
+/**
+ * Serializer for media.
+ */
+function se_media(obj, children) {
+    const mediaAlt = obj.nodes.find(c => {
+        return c.type === 'media_alt'
+    })
```
aiwenar: `c => c.type === 'media_alt'`

src/cnxml/document.js
```diff
@@ -156,17 +156,90 @@ export const LIST_ITEM = block('item', de_list_item, 'list_item', 'item')
...
+
+    const mediaAltHasInlines = mediaAlt ? mediaAlt.nodes.find(n => n.object === 'inline') : false
```
aiwenar: Schema guarantees that there is a `media_alt` block.

src/cnxml/document.js
```diff
@@ -156,17 +156,90 @@ export const LIST_ITEM = block('item', de_list_item, 'list_item', 'item')
...
+
+    return <media alt={mediaAlt ? mediaAlt.text : 'Cannot be empty' }>
```
aiwenar: Don't insert random text into documents.

src/cnxml/document.js
```diff
@@ -156,17 +156,90 @@ export const LIST_ITEM = block('item', de_list_item, 'list_item', 'item')
...
+export const MEDIA = block('media', de_media, 'media', se_media)
+
+export const MEDIA_ALT = {
+    serialize(obj, children) {
+        if (obj.type === 'media_alt') {
+            return <div class="media-alt">{children}</div>
+        }
+        return
+    }
+}
```
aiwenar: Serialization for alt-texts should be defined with `block`.

src/plugins/media/schema.js
```diff
@@ -1,11 +1,27 @@
...
+import { Block, Text } from 'slate'
+
 function normalizeMedia(change, error) {
-    const { code: violation, key, node } = error
+    const { code: violation, node, key } = error
 
     switch (violation) {
+    case 'child_min_invalid':
+        console.log('media child_min_invalid', node.toJS())
+        if (!node.nodes.some(n => n.type && n.type === 'media_alt')) {
```
aiwenar: You can use `error.index` to determine from where the node is missing. When it's 0 then it must be an `image`, when 1 it must be a `media_alt`.

src/plugins/media/schema.js
```diff
@@ -21,24 +37,36 @@ function normalizeMedia(change, error) {
...
-            data: {
-                src: Boolean,
+            media_alt: {
+                nodes: [{
+                    match: [
+                        ...inlines.map(type => ({ type })),
+                        { object: 'text' },
+                    ],
+                }],
+                marks: [],
```
aiwenar: There's no need to specify an empty `marks`.

PiotrKozlowski: `marks: undefined` will allow every mark

src/cnxml/document.js
```diff
@@ -156,17 +156,90 @@ export const LIST_ITEM = block('item', de_list_item, 'list_item', 'item')
...
+
+export const DIV = block('div', de_div, 'div', se_div)
+
+export const MEDIA = block('media', de_media, 'media', se_media)
+
+export const MEDIA_ALT = {
+    serialize(obj, children) {
+        if (obj.type === 'media_alt') {
+            return <div class="media-alt">{children}</div>
+        }
```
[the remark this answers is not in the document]
PiotrKozlowski: It's handled in `media` serializer so we don't have error for not returning anything for `media_alt` blocks.

PiotrKozlowski:
> Also please remove Delete and Backspace handling for now.

They wasn't for me when I was testing, but as you wish :D

Do you have idea how to make `handle-delete-in-media-alt.js` work correctly?

src/cnxml/document.js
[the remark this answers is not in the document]
aiwenar: You mean this?

```js
mediaAltHasInlines ?
                children
            : children.filter(c => c.type !== 'div')
```

This isn't a good solution, as it removes all `div`s regardless of their use. This means that should we decide to add another extensions we would also have to change this code (which we'll probably forget about), and makes it impossible to introduce external extensions (as we planned with translations for text in images).

As for those errors; have you tried returning an empty React fragment (`<></>`) or an empty array?

src/plugins/media/schema.js
```diff
@@ -21,24 +37,36 @@ function normalizeMedia(change, error) {
...
+        blocks: {
+            // TODO: do we actually want to keep multiple versions?
+            media: {
+                nodes: [
+                    { match: { type: 'image' }, min: 1 },
+                    { match: { type: 'media_alt' }, min: 1, max: 1 },
+                ],
+                data: {
+                    alt: Boolean,
+                },
```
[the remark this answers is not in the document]
aiwenar: This doesn't actually mean that we must also keep it in `data`. You can add data entries which are not in schema, they'll just get normalized away at some point (and we can make that normalization move the content to a `media_alt` block if one doesn't exist already).

src/cnxml/document.js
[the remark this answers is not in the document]
PiotrKozlowski:
```
export const MEDIA_ALT = {
    serialize(obj, children) {
        if (obj.type === 'media_alt') {
            if (obj.nodes.some(n => n.object !== 'text')) {
                return <div class="media-alt">{children}</div>
            }
            return <></> // or []
        }
        return
    }
}
```
with
```
/**
 * Serializer for media.
 */
function se_media(obj, children) {
    const mediaAlt = obj.nodes.find(c => c.type === 'media_alt')

    return <media alt={mediaAlt.text}>
        {children}
    </media>
}
```
It doesn't work. We could probably adjust `xml.js` to this case but it seems even more hacky than just specifing class of divs to omit in media serializer.

src/plugins/media/schema.js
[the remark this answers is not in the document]
PiotrKozlowski: It seems like one more, not standard thing which we will have to remember when doing changes or introducing someone to the project.

src/cnxml/document.js
```diff
@@ -175,45 +175,43 @@ function de_media(el, next) {
...
- * Deserializer for divs.
+ * Process data for media_alt nodes.
  */
-function de_div(el, next) {
-    const className = el.getAttribute('class')
-
-    switch (className) {
-    case 'media-alt':
+function de_media_alt(el, next) {
+    if (el.tagName === 'DIV' && el.className === 'media-alt') {
```
aiwenar: You don't need to check for `el.tagName`, `block` has already done this.

PR #61, opened 2019-09-05
Footnotes

PiotrKozlowski:
Add support for footnotes.

~~`{ data: { collapse: boolean } }` is for hiding / showing text.~~

Front: https://github.com/openstax-poland/adaptarr-front/pull/157

### DIFF
--- a/src/plugins/footnote/render.js
+++ b/src/plugins/footnote/render.js
@@ -1,0 +1,11 @@
+// Copyright 2019 OpenStax Poland
+// Licensed under the MIT license. See LICENSE file in the project root for
+// full license text.
+
+import React from 'react'
+
+export default function renderInline({ node, children, attributes }, editor, next) {
+    if (node.type !== 'footnote') return next()
+
+    return <span className="footnote" {...attributes}>{children}</span>
+}
--- a/src/plugins/footnote/schema.js
+++ b/src/plugins/footnote/schema.js
@@ -1,0 +1,38 @@
+// Copyright 2019 OpenStax Poland
+// Licensed under the MIT license. See LICENSE file in the project root for
+// full license text.
+
+function normalizeFootnote(change, error) {
+    const { code, child, node } = error
+
+    switch (code) {
+    case 'child_object_invalid':
+        const text = child.getText()
+        change.removeNodeByKey(child.key)
+        change.insertText(text)
+        break
+
+    /* istanbul ignore next */
+    default:
+        console.warn('Unhandled footnote violation:', code)
+        break
+    }
+}
+
+export default function schema({ marks, inlines }) {
+    return {
+        inlines: {
+            footnote: {
+                nodes: [{
+                    match: [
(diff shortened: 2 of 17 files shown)


test/cnxml/de/inline.js
```diff
@@ -12,7 +12,8 @@ export const input = cnxml`
 links to other elements (<link target-id="f1" />),
 elements in other documents (<link target-id="f1" document="d1" />),
 other <link document="d1">documents</link>,
-and <link url="https://example.test">external links</link>.</para>
+and <link url="https://example.test">external links</link>
+<footnote id="footnote-id">Footnote</footnote>.</para>
```
aiwenar: If you put `<footnote>` before `<link>` you'll modify just one line, and keep this sentence grammatically correct.

test/plugins/footnote/insert.js
```diff
@@ -0,0 +1,21 @@
+/** @jsx h */
+
+export default editor => {
+    editor.insertInline({
+        type: 'footnote',
+        data: { collapse: false },
+        nodes: [{ object: 'text', text: ' ' }],
+    }).moveBackward()
```
aiwenar: When a user inserts a footnote in the editor it should be empty, and test cases should reflect that.

test/util/h.js
```diff
@@ -45,6 +45,7 @@ export default global.h = createHyperscript({
         codeinline: 'code',
         docref: 'docref',
         link: 'link',
+        footnote: 'footnote',
         xref: 'xref',
         term: 'term',
```
aiwenar: This should be alphabetically sorted.

test/plugins/footnote/insert.js
PiotrKozlowski: If it's empty then it will be removed and there is no way to insert cursor in it.
This is way it's added with space inside. Do you see another option to handle this case?

aiwenar: Unfortunately no.

src/plugins/footnote/schema.js
```diff
@@ -0,0 +1,39 @@
...
+        inlines: {
+            footnote: {
+                nodes: [{
+                    match: [
+                        ...inlines.map(type => ({ type })),
+                        { object: 'text' },
+                    ],
+                }],
+                marks: marks.map(type => ({ type })),
+                data,
```
aiwenar: I would prefer schema for data to be part of the UI plugin.

src/plugins/footnote/render.js
```diff
@@ -0,0 +1,11 @@
+// Copyright 2019 OpenStax Poland
+// Licensed under the MIT license. See LICENSE file in the project root for
+// full license text.
+
+import React from 'react'
+
+export default function renderInline({ node, children, attributes }, editor, next) {
+    if (node.type !== 'footnote') return next()
+
+    return <span {...attributes}>{children}</span>
```
aiwenar: All generic elements should have a `className`.

PR #64, opened 2019-09-25
Preformat tag

PiotrKozlowski:
Add support for `<preformat>` element.

We decided to support `<preformat>` only as `block` element since `<pre>` is not allowed in `<p>` anyway.

### DIFF
--- a/src/cnxml/document.js
+++ b/src/cnxml/document.js
@@ -251,6 +251,31 @@ export const MEDIA_ALT = block('div', de_media_alt, 'media_alt', se_media_alt)
 
 export const PARA = block('para', text('paragraph'), 'paragraph', 'para')
 
+/**
+ * Process data for preformat tags.
+ *
+ * Preformat tag may in CNXML may have display="inline" attribute, but since
+ * <pre> tag is block element we support only this variation.
+ */
+function de_preformat(el, next) {
+    return {
+        type: 'preformat',
+        nodes: next(el.childNodes),
+    }
+}
+
+/**
+ * Serializer for preformat.
+ */
+function se_preformat(obj, children) {
+    return <preformat id={obj.key}>
+        {children}
+    </preformat>
+}
+
+export const PREFORMAT = block(
+    'preformat', de_preformat, 'preformat', se_preformat)
+
 export const PROBLEM = block(
     'problem', 'exercise_problem', 'exercise_problem', 'problem')
 
@@ -276,6 +301,7 @@ export const DOCUMENT = [
     MEDIA,
     MEDIA_ALT,
     PARA,
+    PREFORMAT,
     PROBLEM,
     QUOTE,
     SECTION,
--- a/src/plugins/index.js
+++ b/src/plugins/index.js
@@ -11,6 +11,7 @@ import Footnote from './footnote'
 import GlossaryDocument from './glossary'
 import List from './list'
 import Media from './media'
+import Preformat from './preformat'
 import Quotation from './quotation'
 import Section from './section'
 import Term from './term'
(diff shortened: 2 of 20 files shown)


test/plugins/preformat/query-active.js
```diff
@@ -0,0 +1,22 @@
...
+export default editor => {
+    let block = editor.value.startBlock
+
+    let node = block.type === 'preformat' ? block : null
+    should.not.exist(node)
+
+    editor.moveToStartOfNode(editor.value.document.getNode('preformat'))
+    block = editor.value.startBlock
+    node = block.type === 'preformat' ? block : null
+    should.exist(node)
```
aiwenar: If `block` can be `null` or `undefined`, then `block.type` will fail. If it can't then there's no reason to use this elaborate construction with `should.exist` over a simple `block.type.should.equal('preformat')`.

PiotrKozlowski:
> This is almost identical to the code plugin. I'd prefer preformat to be just a variant of a code block, and to reuse for example handlers.

Not sure how to do this since `code` can be also an `inline` and have additional `data` which does not exists for `preformat`.

Handlers however are the same, but I', not sure what would be best option to reuse them. We could have function which will take `block.type` and return proper handlers, but where it should be stored? It feels like if we would keep `preformat` as separate plugin then duplicated code isn't so bad in this case.

If you would like me to somehow merge those two, let me know how you imagine this.

src/cnxml/document.js
```diff
@@ -251,6 +251,33 @@ export const MEDIA_ALT = block('div', de_media_alt, 'media_alt', se_media_alt)
...
+ *
+ * Preformat tag may in CNXML may have display="inline" attribute, but since
+ * <pre> tag is block element we support only this variation.
+ */
+function de_preformat(el, next) {
+    return {
+        object: 'block',
+        type: 'preformat',
+        key: el.getAttribute('id') || undefined,
+        nodes: next(el.childNodes),
```
aiwenar: `object` and `key` are handled by `block`,  you don't need to do it here.

src/plugins/preformat/index.js
```diff
@@ -0,0 +1,29 @@
...
+ *                                   a preformat.
+ */
+export default function Preformat(options={}) {
+    const {
+        inlines = [],
+        marks = [],
+    } = options
+
+    const schema = make_schema({
+        inlines: inlines.concat(ALLOWED_INLINES),
```
aiwenar: You shouldn't concatenate type arrays, as you're likely to get duplicate entries.

src/plugins/preformat/index.js
```diff
@@ -0,0 +1,29 @@
+// Copyright 2019 OpenStax Poland
+// Licensed under the MIT license. See LICENSE file in the project root for
+// full license text.
+
+import onKeyDown from './handlers'
+import renderBlock from './render'
+import make_schema from './schema'
+
+const ALLOWED_INLINES = ['code', 'docref', 'link', 'term', 'xref']
```
aiwenar: We're trying to keep plugins separate, they shouldn't reference each other.

src/plugins/index.js
```diff
@@ -57,9 +60,15 @@ export function TextContent(options={}) {
...
 
+    const preformat = {
+        inlines: ['code', 'docref', 'link', 'term', 'xref'],
+        ...options.preformat || {},
+    }
+
     return [
         Code(options.code),
         Footnote({ marks }),
+        Preformat({ marks, preformat }),
```
aiwenar: `Preformat` takes an object with `marks` and `inlines` properties, not `marks` and `preformat`. You've probably meant 

```suggestion
        Preformat({ marks, ...preformat }),
```

PR #65, opened 2019-09-26
Define default inlines which are allowed in footnote

PiotrKozlowski:


aiwenar:
Footnotes should allow the same marks and inlines as all other text containers.

PiotrKozlowski:
By TextContainers you mean `Code`, `Term`, `Text` and `XReference` pluggins?
`Code`, `Term` and `Footnote` accepts inlines passed by options.
`Text` accepts all inlines.

Or do you mean that footnotes should accept all inlines by default?

PiotrKozlowski:
Moved to: https://github.com/openstax-poland/cnx-designer/pull/66

PR #69, opened 2019-10-15
Fix for cursor.delete()

PiotrKozlowski:
`cursor.delete()` is async

If `cursor => {...}` is not `async` I'm getting this error while for. ex. saving document:
`TypeError: f(...) is undefined` for:
```
function iterate(store, ...args) {
    const f = args.pop()

    return new Promise((resolve, reject) => {
        const req = store.openCursor(...args)
        req.onerror = err => reject(err)
        req.onsuccess = event => {
            const cursor = event.target.result
            if (cursor) {
                f(cursor, cursor.value, reject).catch(reject)
            } else {
                resolve()
            }
        }
    })
}
```
from `src/persistance/index.js`

Should we wrap `cursor.delete()` with `promisify`? The `await` statement is acctually not necessary. Using only `async` is enough to make it work.

### DIFF
--- a/src/persistence/index.js
+++ b/src/persistence/index.js
@@ -151,8 +151,8 @@ export class PersistDB {
 
             if (index && keys) {
                 for (const key of keys) {
-                    await iterate(store.index(index), key, cursor => {
-                        cursor.delete()
+                    await iterate(store.index(index), key, async cursor => {
+                        await cursor.delete()
                         cursor.continue()
                     })
                 }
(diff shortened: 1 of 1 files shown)


aiwenar:
I did wonder why those callbacks were marked as `async`. 

IndexedDB methods don't return promises, so `await cursor.delete()` is wrong; `await promisify(cursor.delete())` could be OK, but seeing that it wasn't needed so far, I think the best solution would be to alter `iterate` so that it doesn't treat the callback as a promise (see #73).

PR #70, opened 2019-10-15
Use new context API

PiotrKozlowski:


src/api/storage.js
```diff
@@ -2,14 +2,18 @@
...
 /* eslint-disable no-unused-vars */
 
-export default class Storage {
+export class Storage {
     /**
      * Return an URL for a given media file.
      */
     mediaUrl(name) {
         throw new Error("not implemented")
     }
```
aiwenar: This class was only meant to represent an interface, and was only implemented as a class to make `PropTypes.instanceOf` work. Since after this change it will be the default value, this method must not throw but return a usable value (and perhaps log a warning).

src/api/storage.js
```diff
@@ -2,14 +2,18 @@
...
 }
+
+export const StorageContext = React.createContext({ storage: new Storage() })
```
aiwenar: This should not be an object.

PiotrKozlowski:
> I'm wondering whether exporting StorageContext as Storage.Context wouldn't be nicer.

I think `NameContext` is standard naming for contexts

src/api/storage.js
```diff
@@ -5,15 +5,15 @@
...
 
 export class Storage {
     /**
      * Return an URL for a given media file.
      */
     mediaUrl(name) {
-        throw new Error("not implemented")
+        console.warn('You are using default Storage class.')
+        return name
     }
```
aiwenar: _You are using default Storage class_ suggests, that this is an acceptable/expected behaviour. It is not; this class should never be used and was only provided as a superclass (and now as a fallback). This message should be clear that this is a bug and you should fix it.

aiwenar:
I've just realised, that there is now no reason to use a class at all. Since all this class is is a single function, we could just use an actual function.

PiotrKozlowski:
> I've just realised, that there is now no reason to use a class at all. Since all this class is is a single function, we could just use an actual function.

Do I understand that we should remove `Storage` class, `StorageContext` and add `mediaUrl` function should be retrieved from `Image` plugin's options?

PR #72, opened 2019-10-16
Allow only specified children in list_item

PiotrKozlowski:
Currently all `block` nodes are allowed in `list_item`
This pr is changing this so now only specified types will be allowed.
It's done by passing `typesChildren` to `List` plugin options (name of prop is consistent with other plugin's option).

Tests are working correctly, but when I drag for example `definition` node into list item, it's not normalized. I can also refresh page and this incorrect item still will be there.
obraz

### DIFF
--- a/src/plugins/list/index.js
+++ b/src/plugins/list/index.js
@@ -1,4 +1,4 @@
-// Copyright 2018 OpenStax Poland
+// Copyright 2019 OpenStax Poland
 // Licensed under the MIT license. See LICENSE file in the project root for
 // full license text.
 
@@ -6,9 +6,17 @@ import EditList from 'slate-edit-list'
 
 import * as commands from './commands'
 import renderBlock from './render'
+import make_schema from './schema'
 
+/**
+ * @param {EditList~OptionsFormat} options - Object with options for EditList
+ *                                           plugin.
+ * @param {string[]} options.content - List of node types which
+ *                                     are allowed inside list items.
+ */
 export default function List(options) {
     const list = EditList(options)
-    const plugin = { commands, renderBlock }
+    const schema = make_schema(options)
+    const plugin = { commands, renderBlock, schema }
     return [plugin, list]
 }
--- a/src/plugins/list/schema.js
+++ b/src/plugins/list/schema.js
@@ -1,0 +1,59 @@
+// Copyright 2019 OpenStax Poland
+// Licensed under the MIT license. See LICENSE file in the project root for
+// full license text.
+
+function normalizeList(change, error) {
+    const { child, code } = error
+
+    switch (code) {
+    case 'child_type_invalid': {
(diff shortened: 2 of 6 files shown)


src/plugins/list/schema.js
```diff
@@ -0,0 +1,40 @@
...
+function normalizeListItem(change, error) {
+    const { child, code } = error
+
+    switch (code) {
+    case 'child_type_invalid': {
+        if (child.object === 'text') {
+            change.wrapNodeByKey(child.key, Block.create('paragraph'))
+            break
+        }
+        change.removeNodeByKey(child.key)
```
aiwenar: Invalid children should be unwrapped.

src/plugins/list/schema.js
```diff
@@ -0,0 +1,40 @@
...
+
+import { Block } from 'slate'
+
+function normalizeListItem(change, error) {
+    const { child, code } = error
+
+    switch (code) {
+    case 'child_type_invalid': {
+        if (child.object === 'text') {
+            change.wrapNodeByKey(child.key, Block.create('paragraph'))
```
aiwenar: You're assuming `paragraph` is a valid child, but also allow configuring the plugin such that it isn't.

You can also write this as `change.wrapNodeByKey(child.key, 'paragraph')`.

src/plugins/list/index.js
```diff
@@ -1,14 +1,22 @@
...
 import * as commands from './commands'
 import renderBlock from './render'
+import make_schema from './schema'
 
+/**
+ * @param {EditList~OptionsFormat} options - Object with options for EditList
+ *                                           plugin.
+ * @param {string[]} options.typesChildren - List of node types which
+ *                                           are allowed inside list items.
+ */
```
aiwenar: To keep consistency with other plugins this should be named `content`.

src/plugins/list/schema.js
PiotrKozlowski: > You can also write this as change.wrapNodeByKey(child.key, 'paragraph').

This would cause: `Error: `Node.create` only accepts objects or nodes but you passed it: paragraph`, but I'l change it to `{ object: 'block', type: 'paragraph' }`

aiwenar: Ah right, `wrapNodeByKey` is just a wrapper around either `wrapBlockByKey` or `wrapInlineByKey`, and thus can't distinguish between those cases if it's provided a string. `change.wrapBlockByKey(child.key, 'paragraph')` is the correct short version.

aiwenar: > You're assuming paragraph is a valid child, but also allow configuring the plugin such that it isn't.

Turns out slate-edit-list on which this plugin is based already assumes that `paragraph` exists, so I retract this concern.

PR #75, opened 2019-11-05
Add support for foreign tag

PiotrKozlowski:
Add support for foreign tag.

Additionally add `xml:lang` attribute - I've assumed that since we support importing cnxml we do not want to make this attribute required.

Things to consider:
Should we validate `data.lang` against languages codes from: https://github.com/openstax-poland/adaptarr-front/blob/master/src/locale/data.json ?
On the front end side user will be able to choose only from those codes.

### DIFF
--- a/test/plugins/foreign/insert-nested-inline.js
+++ b/test/plugins/foreign/insert-nested-inline.js
@@ -1,0 +1,27 @@
+/** @jsx h */
+
+export default editor => editor.wrapInline({ type: 'term' })
+
+// There is probably a bug in slate which causes selection to
+// be moved at the end for term.
+export const checkSelection = false
+
+export const input = <value>
+    <document>
+        <p>
+            <text/>
+            <foreign>Foreign <anchor/>term<focus/> text</foreign>
+            <text/>
+        </p>
+    </document>
+</value>
+
+export const output = <value>
+    <document>
+        <p>
+            <text/>
+            <foreign>Foreign <term><anchor/>term</term><focus/> text</foreign>
+            <text/>
+        </p>
+    </document>
+</value>
--- a/test/plugins/foreign/schema-bare-foreign.js
+++ b/test/plugins/foreign/schema-bare-foreign.js
@@ -1,0 +1,15 @@
+/** @jsx h */
+
+export default editor => editor.normalize()
+
+export const input = <value>
+    <document>
+        <foreign>Foreign</foreign>
+    </document>
+</value>
+
+export const output = <value>
+    <document>
+        <p><text/></p>
+    </document>
+</value>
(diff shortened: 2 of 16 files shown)


test/plugins/foreign/insert-nested-inline.js
```diff
@@ -0,0 +1,23 @@
...
+            <text/>
+        </p>
+    </document>
+</value>
+
+export const output = <value>
+    <document>
+        <p>
+            <text/>
+            <foreign>Foreign <term>term</term><cursor/> text</foreign>
```
aiwenar: Selection should stay around the term.

test/plugins/foreign/schema-bare-foreign.js
```diff
@@ -0,0 +1,15 @@
...
+    <document>
+        <foreign>Foreign</foreign>
+    </document>
+</value>
+
+export const output = <value>
+    <document>
+        <p><text/></p>
+    </document>
+</value>
```
aiwenar: This term should be moved into the paragraph, not removed.

PiotrKozlowski:
After serializing `data.lang` with:

I get:
```
2) CNXML
       se
         inline:
     AssertionError: At /content[0]/para[19]/foreign/@{http://www.w3.org/XML/1998/namespace}lang: unexpected attribute
      at new HtmlError (test/util/compareHtml.js:4:50)
      at compareHtml (test/util/compareHtml.js:58:23)
      at compareHtml (test/util/compareHtml.js:104:13)
      at compareHtml (test/util/compareHtml.js:104:13)
      at fn (test/cnxml/index.js:64:9)
      at Context.runTest (test/util/fixtures.js:37:17)
```

When I deserialize `lang` with:
```
data: {
            lang: el.getAttributeNS(NAMESPACES.xml, 'lang'),
        },
```

I get:
```
1) CNXML
       de
         inline:

      AssertionError: expected { Object (object, data, ...) } to equal { Object (object, data, ...) }
      + expected - actual

                 "text": ", "
               }
               {
                 "data": {
      -            "lang": [null]
      +            "lang": "pl"
                 }
                 "nodes": [
                   {
                     "marks": []
      
      at fn (test/cnxml/index.js:37:21)
      at Context.runTest (test/util/fixtures.js:37:17)
```

test/plugins/foreign/schema-bare-foreign.js
PiotrKozlowski: There is only `normalizeDocument child_min_invalid` returned as an error for this test.
It might be because `document` can contain only `block` elements. 
What would you like me to do with this test?

test/plugins/foreign/insert-nested-inline.js
PiotrKozlowski: This is current behavior of slate. Would you like me to remove this test, comment it out or do something else?

aiwenar:
You get those errors because you forgot to adjust tests; they still use `<term lang="..">` rather than `<term xml:lang="..">`.

PR #77, opened 2019-11-12
Add support for audio and video elements

PiotrKozlowski:
Things to consider:
Currently `media` is allowed only inside `figure`.
Should we allow `media` also outside `figure`?

https://github.com/openstax-poland/adaptarr-front/issues/222

### DIFF
--- a/src/cnxml/document.js
+++ b/src/cnxml/document.js
@@ -133,13 +133,18 @@ export const FIGURE = block(
  * Process data for images.
  */
 function de_image(el) {
+    const data = {
+        mime: el.getAttribute('mime-type'),
+        src: el.getAttribute('src'),
+    }
+    if (el.hasAttribute('for')) {
+        data.for = el.getAttribute('for')
+    }
+
     return {
         type: 'image',
         isVoid: true,
-        data: {
-            src: el.getAttribute('src'),
-            mime: el.getAttribute('mime-type'),
-        },
+        data,
     }
 }
 
@@ -147,15 +152,78 @@ function de_image(el) {
  * Serializer for images.
  */
 function se_image(obj) {
-    const attrs = {
-        src: obj.data.get('src'),
-        'mime-type': obj.data.get('mime'),
-    }
+    const attrs = obj.data.toJS()
+    attrs['mime-type'] = obj.data.get('mime')
+    delete attrs.mime
 
     return <image {...attrs} />
 }
 
 export const IMAGE = block('image', de_image, 'image', se_image)
+
+/**
+ * Process data for audio.
+ */
+function de_audio(el) {
+    const data = {
+        mime: el.getAttribute('mime-type'),
+        src: el.getAttribute('src'),
+    }
+    if (el.hasAttribute('for')) {
+        data.for = el.getAttribute('for')
+    }
+
+    return {
+        type: 'audio',
+        isVoid: true,
+        data,
+    }
+}
+
+/**
+ * Serializer for audio.
+ */
+function se_audio(obj) {
+    const attrs = obj.data.toJS()
+    attrs['mime-type'] = obj.data.get('mime')
+    delete attrs.mime
+
+    return <audio {...attrs} />
+}
+
+export const AUDIO = block('audio', de_audio, 'audio', se_audio)
+
+/**
+ * Process data for video.
+ */
+function de_video(el) {
+    const data = {
+        mime: el.getAttribute('mime-type'),
+        src: el.getAttribute('src'),
+    }
+    if (el.hasAttribute('for')) {
+        data.for = el.getAttribute('for')
+    }
+
+    return {
+        type: 'video',
+        isVoid: true,
+        data,
+    }
+}
+
+/**
+ * Serializer for video.
+ */
+function se_video(obj) {
+    const attrs = obj.data.toJS()
+    attrs['mime-type'] = obj.data.get('mime')
+    delete attrs.mime
+
+    return <video {...attrs} />
+}
+
+export const VIDEO = block('video', de_video, 'video', se_video)
 
 /**
  * Process data for list nodes.
@@ -347,6 +415,7 @@ export const TITLE = block('title', text('title'), 'title', 'title')
 export const DOCUMENT = [
     ADMONITION,
     ALT_TEXT,
+    AUDIO,
     FIGURE_CAPTION,
     CODE,
     COMMENTARY,
@@ -368,4 +437,5 @@ export const DOCUMENT = [
     SOLUTION,
     STATEMENT,
     TITLE,
+    VIDEO,
 ]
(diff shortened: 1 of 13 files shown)


src/cnxml/util.js
```diff
@@ -361,6 +361,42 @@ export function loadClasses(el) {
...
+            attrs[attr] = val.join(' ')
+            continue
+        } else if (attr === 'mime') {
+            attrs['mime-type'] = val
+            continue
+        }
+        attrs[attr] = val
+    }
+    return attrs
+}
```
aiwenar: I'm not convinced we should indiscriminately load all attributes, and attributes should be loaded as their proper types, not as strings.

src/plugins/media/schema.js
```diff
@@ -42,10 +42,16 @@ function normalizeMedia(change, error) {
 export default function schema({ inlines }) {
     return {
         blocks: {
-            // TODO: do we actually want to keep multiple versions?
```
aiwenar: This is still an unresolved question.

test/plugins/figure/insert-additional-image.js
```diff
@@ -0,0 +1,51 @@
...
+                    key="first"
+                    src="first.png"
+                    mime="image/png"
+                    for="online"
+                    >
+                    <text><cursor/></text>
+                </img>
+                <img src="second.jpg" mime="image/jpeg" for="pdf">
+                    <text><cursor/></text>
+                </img>
```
aiwenar: We shouldn't test support for alternative media files until we decide on how to handle them.

PiotrKozlowski: I thought this was clear? Supporting `for` attribute from `cnxml` seems enough for those.

src/plugins/media/schema.js
PiotrKozlowski: This is about `for`, right? It seems like using this attribute is right way to do this

src/cnxml/util.js
PiotrKozlowski: I've added this to not repeat myself when reading attrs for media files and in the future for elements related to tables since there will be many of the same attrs.
I agree that elements should be loaded as proper types, not strings.

What is your suggestion for handling this overall?
- Duplicate converting attrs for each media file
- Leave it as it is + convert attrs to proper types + omit unwanted attrs?
- Something else?

PiotrKozlowski:
Build is failing at error which was fixed with another pr.
All requested changes were provided except of removing `for`. I think it is good enough to just keep it as it is. Additional validation is not required since we can't just remove `media` element because for ex. user added only one media and set `for="pdf"`. It is probably mistake, but there is nothing we could do in this case from `cnx-designer` repo

src/cnxml/document.js
```diff
@@ -133,18 +131,40 @@ export const FIGURE = block(
...
+    }
+    if (el.hasAttribute('longdesc')) {
+        data.longdesc = el.getAttribute('longdesc')
+    }
+    if (el.hasAttribute('thumbnail')) {
+        data.thumbnail = el.getAttribute('thumbnail')
+    }
+    if (el.hasAttribute('width')) {
+        data.width = Number(el.getAttribute('width'))
+    }
```
aiwenar: I'm not convinced we want to support all those attributes. I'm not sure whether the modern online/pdf pipeline even makes use of width, height, or thumbnail. As for longdesc, I have never seen it used in OpenStax's books and I don't know how it's supposed to be used / how it works. A similar [HTML property](https://developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement/longDesc) is deprecated.

src/cnxml/document.js
```diff
@@ -155,18 +175,46 @@ export const IMAGE = block('image', de_image, 'image', se_image)
...
+    if (el.hasAttribute('loop')) {
+        data.autoplay = el.getAttribute('loop') === 'true'
+    }
+    if (el.hasAttribute('standby')) {
+        data.standby = el.getAttribute('standby')
+    }
+    if (el.hasAttribute('volume')) {
+        data.volume = Number(el.getAttribute('volume'))
+    }
```
aiwenar: I'm not convinced we want to support all those attributes (except for `for` but I could not not select it). We should first find examples in OpenStax's books (preferably, third party books on CNX would be find) of how those attributes are used and how they are transformed into HTML. Also see discussion of longdesc above.

src/cnxml/document.js
```diff
@@ -155,18 +175,46 @@ export const IMAGE = block('image', de_image, 'image', se_image)
...
+    if (el.hasAttribute('for')) {
+        data.for = el.getAttribute('for')
+    }
+    if (el.hasAttribute('longdesc')) {
+        data.longdesc = el.getAttribute('longdesc')
+    }
+    if (el.hasAttribute('loop')) {
+        data.autoplay = el.getAttribute('loop') === 'true'
+    }
```
aiwenar: This should probably be `data.loop = ...`.

src/cnxml/document.js
```diff
@@ -177,18 +225,52 @@ export const AUDIO = block('audio', de_audio, 'audio', se_audio)
...
+    if (el.hasAttribute('standby')) {
+        data.standby = el.getAttribute('standby')
+    }
+    if (el.hasAttribute('width')) {
+        data.width = Number(el.getAttribute('width'))
+    }
+    if (el.hasAttribute('volume')) {
+        data.volume = Number(el.getAttribute('volume'))
+    }
```
aiwenar: See discussion above.

src/cnxml/document.js
PiotrKozlowski: I think `width` and `height` may be useful even if they are not used by OS.
About `thumbnail` and `longdesc` - I've added those just because they exists in `cnxml` and I'm not sure if we should remove them if they were present when loading document

aiwenar: We shouldn't support something just because it may be useful. Also, the seem too much like formatting to me, and we explicitly don't support formatting.

As for removing attributes; we already don't guarantee that the editor will preserve anything, and in fact it often modifies documents, so I don't think that's a problem.

PiotrKozlowski: > we already don't guarantee that the editor will preserve anything

I know, but I don't think this is good approach. I would like it to be more like: preserve anything which we can and allow editing things which we want to be edited

Let me know which attributes you would like to support from:
https://legacy.cnx.org/eip-help/image
https://legacy.cnx.org/eip-help/audio
https://legacy.cnx.org/eip-help/video

PiotrKozlowski: Decision: as little as possible

PiotrKozlowski: According to: https://katalysteducation.slack.com/archives/GCSUE8C01/p1576223516003500
support those tags which can be used in cnxml and disable those which we does not want to support in front end

PR #78, opened 2019-11-14
Lint pre-commit and test + build pre-push

PiotrKozlowski:
Lint pre-commit and test + build pre-push

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,9 @@
   "main": "dist/index.cjs.js",
   "module": "dist/index.es.js",
   "private": true,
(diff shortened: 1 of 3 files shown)


aiwenar:
I'm against including a dependency many users (including me) won't use, especially when it introduces additional time overhead to performing basic actions when enabled. I also don't believe we need it; we have a CI which runs those tests, rarely push to the master, and (hopefully) have proper discipline of running tests and lints before pushing commits. And if someone finds such git hooks useful they can easily set them up themselves.

PR #80, opened 2019-11-21
Introduce tables

PiotrKozlowski:
Basic version of tables:
- add columns
- add rows
- de/serialize all types of table related blocks and attributes

TODO:
- [x] Add tests (two tests are failing, no idea why :/)
- [x] Handle invalid children
- [x] Handle invalid attributes (ex. convert string `"2"` to number `2`
- [x] Disallow removing table cells (`entry`) with <kbd>Backspace</kbd> /  <kbd>Delete</kbd>
- [x] Disallow adding table cells (`entry`) with <kbd>Enter</kbd>

In this pr we introduce tables. All already created tables should be deserialized properly. All values from attributes should be retrieved and kept.

In next pr, when we decide how to handle merging cells we will probably remove some of unnecessary blocks and attributes.


aiwenar:
Will be re-done for Slate 0.5x after update.

PR #86, opened 2019-12-03
Implement <rule>

aiwenar:


src/plugins/rule/commands.js
```diff
@@ -0,0 +1,196 @@
...
+ * Create a new child element, insert it into current rule, and collapse
+ * selection into it.
+ *
+ * Fails if no rule is selected.
+ *
+ * @param {Slate~Editor} editor
+ * @param {string} type         type of element to insert
+ * @param {string} ...previous  type of elements before the one to insert
+ */
+function insertInRule(editor, type, ...previous) {
```
PiotrKozlowski: It would be easier to read if previous would be just an array of string since it will be converted to an array anyway

src/plugins/rule/schema.js
```diff
@@ -0,0 +1,143 @@
...
+    switch (code) {
+    // A child of different type was expected.
+    case 'child_type_invalid': {
+        // A text child could only have been added to an exercise by Slate in
+        // event that all other children were removed. In such case we just want
+        // to drop the exercise.
+        if (Text.isText(child)) {
+            editor.removeNodeByKey(node.key)
+            return
+        }
```
PiotrKozlowski: It can also be added by importing wrong cnxml. Text should be either moved outside of parent or wrapped with paragraph and then wrapped with statement / proof / example

src/plugins/rule/schema.js
```diff
@@ -0,0 +1,143 @@
...
+        // A non-empty rule without a statement should get an empty statement.
+        if (index === 0) {
```
PiotrKozlowski: Since there can be a `title` at 0 index, `statement` can be missing also with `index === 1`, but I think it will be called with `child_min_invalid` error

src/plugins/rule/schema.js
```diff
@@ -0,0 +1,143 @@
...
+        // Otherwise (invalid child after the statement but not at the end of
+        // rule) we fold the invalid child into previous node.
+        const previous = node.nodes.get(index - 1)
+        editor.moveNodeByKey(child.key, previous.key, previous.nodes.size)
+        break
+    }
+
+    case 'child_min_invalid':
+        // A non-empty rule without a statement should get an empty statement.
+        if (index === 0) {
```
PiotrKozlowski: `statement` is only child which can cause this error so there is no need to check `index`
I would suggest to first check if `rule_statement` is existing, but in wrong place and then if yes -> move it to correct position and if not then create new `statement`

src/plugins/rule/schema.js
```diff
@@ -0,0 +1,143 @@
...
+        break
+
+    /* istanbul ignore next */
+    default:
+        console.warn('Unhandled rule content violation:', code)
+        break
+    }
+}
+
+export const RULE_PARENT = ['document', 'section']
```
PiotrKozlowski: Unused

src/plugins/rule/schema.js
```diff
@@ -0,0 +1,143 @@
...
+export default function schema(options) {
+    const content = {
+        match: options.content.map(type => ({ type })),
+        min: 1,
+    }
+
+    return {
+        blocks: {
+            rule: {
+                parent: [{ object: 'document' }, { type: 'section' }],
```
PiotrKozlowski: If we want to restrict parents then it should be configurable by options

src/plugins/rule/xref.js
```diff
@@ -0,0 +1,31 @@
...
+        return `Statement ${rule}.${statement}`
+    }
+
+    case 'rule_proof': {
+        const rule = counters.get('rule')
+        const proof = counters.get('rule_proof')
+        return `Proof ${rule}.${proof}`
+    }
+
+    case 'rule_commentary': {
```
PiotrKozlowski: `rule_example`?

test/plugins/rule/insert-second-proof.js
```diff
@@ -0,0 +1,38 @@
...
+    <document>
+        <rule type="rule">
+            <statement>
+                <p>Statement</p>
+            </statement>
+            <proof>
+                <p>Proof</p>
+            </proof>
+            <proof>
+                <p><text><cursor/></text></p>
```
PiotrKozlowski: `<cursor/>` should be enough

test/plugins/rule/handle-enter-in-example.js
```diff
@@ -0,0 +1,40 @@
...
+                <p>Statement</p>
+            </statement>
+            <ruleexample>
+                <p>Ex</p>
+            </ruleexample>
+            <ruleexample>
+                <p>ample</p>
+            </ruleexample>
+        </rule>
+        <p><text><cursor/></text></p>
```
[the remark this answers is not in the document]
aiwenar: `<cursor/>` only does anything when placed in a text node, and slate-hyperscript doesn't insert any by itself, hence why `<text>` here is necessary.

test/plugins/rule/insert-invalid-parent.js
```diff
@@ -0,0 +1,21 @@
...
+
+export const output = <value>
+    <document>
+        <rule type="rule">
+            <statement>
+                <p>A note<cursor/></p>
+            </statement>
+        </rule>
+    </document>
+</value>
```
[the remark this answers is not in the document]
aiwenar: Event if it's blocked there's no harm in making sure nothing weird happens in case something breaks and the user is able to put it there.

test/plugins/rule/schema-bad-content.js
```diff
@@ -0,0 +1,42 @@
+/** @jsx h */
+
+export default editor => editor.normalize()
+
+export const input = <value>
+    <document>
+        <rule type="rule">Bare text</rule>
```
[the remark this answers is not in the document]
aiwenar: See rationale in https://github.com/openstax-poland/cnx-designer/blob/rule/src/plugins/rule/schema.js#L15-L17.

src/plugins/rule/schema.js
aiwenar: Such CNXML wouldn't pass validation (since text isn't allowed in `<rule>`), so I don't think it's a concern.

src/plugins/rule/schema.js
aiwenar: You're right, that case wasn't handled. I've fixed it and added a new test case.

src/plugins/rule/schema.js
aiwenar: https://github.com/openstax-poland/cnx-designer/blob/rule/src/plugins/rule/commands.js#L7

test/plugins/rule/handle-enter-in-example.js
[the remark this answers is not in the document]
PiotrKozlowski: It is working without `text` and it is written as so in other tests

src/plugins/rule/schema.js
PiotrKozlowski: When someone imports cnxml when editing draft it is saved correctly on server even if there were errors when validating

src/plugins/rule/schema.js
PiotrKozlowski: Ex. case like this:
```
export const input = <value>
    <document>
        <rule type="rule">
            <proof>
                <p>Missing proof</p>
            </proof>
            <statement>
                <p><text/></p>
            </statement>
        </rule>
    </document>
</value>
```
Will also trigger `child_min_invalid` and we shouldn't add new statement but just move this one to correct position

test/plugins/rule/handle-enter-in-example.js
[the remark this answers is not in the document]
aiwenar: Interesting, this must have been changed since I wrote tests for examples on which those are based (they still wrap cursor in text if the node would otherwise be empty).

src/plugins/rule/schema.js
PiotrKozlowski: I would say then, that better place for this would be in `index.js` with a comment or just in `commands.js`

src/plugins/rule/queries.js
```diff
@@ -0,0 +1,44 @@
...
+            rule: 'enter',
+            rule_statement: 'enclose',
+            rule_proof: 'enclose',
+            rule_example: 'enclose',
+        },
+        rule_statement: {
+            rule_statement: 'enter',
+        },
+        rule_proof: {
+            rule_statement: 'enter',
```
PiotrKozlowski: `rule_proof`

src/plugins/rule/queries.js
```diff
@@ -0,0 +1,44 @@
...
+        },
+        rule_example: {
+            rule_statement: 'enter',
```
PiotrKozlowski: `rule_example`

src/plugins/rule/schema.js
aiwenar: Then we probably should also change other plugins (e.g. exercise) which also restrict their parents. For now I'd leave it as is; we're about to start migration to Slate 0.5X which would require schema to be largely rewritten, so I'd prefer not to invest too much time in this old schema.

src/plugins/rule/schema.js
aiwenar: I still prefer to make sure that the error is relevant. As for statements out of order, I've added a separate check and test for that case.

src/plugins/rule/commands.js
aiwenar: While I agree that this is somewhat strange, there's a good reason for this. `type` and `...previous` when considered together in this order name (in reverse) elements which must appear before the element to be inserted (`type` is also considered a part of `previous` because the newly added node will be after all previous nodes of that type). This could be represented as a single array (i.e. `insertInRule(editor: Editor, previous: string[])`, with TS types added for clarity), but that doesn't signify that the array must be non-empty. To put it differently, this code will generate a compile time warning:

```js
function insertInRule(editor, type, ...previous) { ... }

insertInRule(editor)
```

while this code won't:

```js
function insertInRule(editor, previous_and_type) { ... }

insertInRule(editor, [])
```

As for why it's a spread array rather than a normal array, I think `insertInRule(editor, 'a', 'b', 'c')` conveys the intent better than `insertInRule(editor, 'a', ['b', 'c'])`, when `'a'`, `'b'`, and `'c'` should in fact be considered a single array.

src/plugins/rule/schema.js
aiwenar: Such cases should be handled by the CNXML loader, because as this case shows it's impossible to have the same set of normalization rules handle both editing errors and import errors. And while it is currently impossible/not viable to normalize imports differently (it will be after migration to 0.5X, as `slate-html-serializer` is deprecated and we'll have to replace it with a custom solution, which I wanted to do for a long time), I still believe that we should focus on editing behaviour.

src/plugins/rule/schema.js
aiwenar: What nodes are allowed as parents is still part of the schema, even if this particular variable is not actually used in that file.

PiotrKozlowski: It feels more like this should be passed as param to `commands` rather than hardcoded in different place

src/plugins/rule/schema.js
PiotrKozlowski: Ok, this argument is good enough for me :D

PR #87, opened 2020-01-28
Add data.reference to definition_term

PiotrKozlowski:
https://github.com/openstax-poland/adaptarr-front/issues/290

### DIFF
--- a/src/cnxml/glossary.js
+++ b/src/cnxml/glossary.js
@@ -53,13 +53,43 @@ function de_term(el, next) {
         return undefined
     }
 
+    const data = {}
+
+    const reference = el.getAttributeNS(
+        'http://katalysteducation.org/cmlnle/1.0', 'reference')
+    if (
+        reference != null
+        && el.parentElement
+        && el.parentElement.tagName === 'definition'
+    ) {
+        data.reference = reference
+    }
+
     return {
         type: 'definition_term',
+        data,
         nodes: next(el.childNodes),
     }
 }
 
-export const DEFINITION_TERM = block('term', de_term, 'definition_term', 'term')
+/**
+ * Serializer for terms.
+ */
+function se_term(obj, children) {
+    const attrs = {}
+
+    const reference = obj.data.get('reference')
+    if (reference && reference !== obj.text) {
+        attrs.cmlnleReference = reference
+    }
+
+    return <term {...attrs}>
+        {children}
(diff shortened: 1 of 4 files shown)


test/cnxml/se/admonition-with-glossary-and-definition.js
```diff
@@ -42,7 +44,7 @@ export const output = `
     </content>
     <glossary>
         <definition>
-            <term>Term</term>
+            <term xmlns:cmlnle="http://katalysteducation.org/cmlnle/1.0" cmlnle:reference="other value">Term</term>
```
aiwenar: I'd prefer the namespace to be set on `<document>` (this won't affect tests since CNXML comparison is done on DOM element, not strings), as it is clearer.

test/cnxml/se/admonition-with-glossary-and-definition.js
```diff
@@ -1,5 +1,7 @@
 /** @jsx h */
 
+/* eslint-disable max-len */
```
aiwenar: Please reformat this test so that this isn't required.

test/cnxml/de/admonition-with-glossary.js
```diff
@@ -2,6 +2,8 @@
 
 import { List } from 'immutable'
 
+/* eslint-disable max-len */
```
aiwenar: Please reformat this test so that this isn't necessary.

src/cnxml/glossary.js
```diff
@@ -53,13 +53,38 @@ function de_term(el, next) {
         return undefined
     }
 
+    const data = {}
+
+    if (el.parentElement && el.parentElement.tagName === 'definition') {
+        data.reference = el.getAttributeNS(
+            'http://katalysteducation.org/cmlnle/1.0', 'reference')
+    }
```
aiwenar: Id suggest also checking for presence:

```js
const reference = el.getAttributeNS('http://katalysteducation.org/cmlnle/1.0', 'reference')
if (reference != null && el.parentElement && el.parentElement.tagName === 'definition') {
    data.reference = reference
}
```
this way `data.reference` will only be set if `cmlnle:reference` is actually present, and we can avoid adding `reference={null}` to all tests.

MartaSztuka:
Screenshot from 2020-01-30 10-44-26

PR #88, opened 2020-01-29
Ensure `<foreign>` is always around `<term>`, not the other way around

aiwenar:


MartaSztuka:
tested
Screenshot from 2020-01-30 10-21-42

PR #97, opened 2020-06-30
Lint using new TypeScript configs from eslint-config-openstax-poland

aiwenar:
Requires openstax-poland/eslint-config-openstax-poland#4.

### DIFF
--- a/src/interfaces/classes.ts
+++ b/src/interfaces/classes.ts
@@ -14,10 +14,10 @@ const WHITESPACE = /[ \t\n]/
 
 export const WithClasses = {
     /** Check if value of unknown type has classes */
-    hasClasses(value: any): value is WithClasses {
+    hasClasses(value: unknown): value is WithClasses {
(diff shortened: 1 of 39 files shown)


src/interfaces/classes.ts
```diff
@@ -14,10 +14,10 @@ const WHITESPACE = /[ \t\n]/
 
 export const WithClasses = {
     /** Check if value of unknown type has classes */
-    hasClasses(value: any): value is WithClasses {
+    hasClasses(value: unknown): value is WithClasses {
         return typeof value === 'object'
-            && Array.isArray(value.classes)
-            && typeof value.classes[0] === 'string'
+            && Array.isArray((value as WithClasses).classes)
+            && typeof (value as WithClasses).classes[0] === 'string'
```
PiotrKozlowski: This is why I like using `any` in some situations. Currently, you have add `as WithClasses` just to check if this object has a property which seems odd to me

PR #104, opened 2020-09-24
Allow foregins in definition terms and return undefined if deserializ…

PiotrKozlowski:
Allow foregins in definition terms and return undefined if deserialize function returns undefined

### DIFF
--- a/src/cnxml/glossary.js
+++ b/src/cnxml/glossary.js
@@ -49,7 +49,10 @@ export const SEEALSO = block(
  * Process data for terms.
  */
 function de_term(el, next) {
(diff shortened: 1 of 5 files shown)


MartaSztuka:
Tested on Steging
Screenshot from 2020-09-29 14-09-33

PiotrKozlowski:
@MartaSztuka This is not a `foreign` inside `definition term`:
> definition term is a text in Glossary after "Definition for:"

Please also test this in normal editing mode

MartaSztuka:
@PiotrKozlowski This is foreign in definition term, in glossary, saved and refreshed
Screenshot from 2020-09-30 12-10-39

PR #105, opened 2020-09-24
Handle classnames for document and glossary elements

PiotrKozlowski:


MartaSztuka:
Tested on Staging
Screenshot from 2020-09-29 14-50-32

PR #110, opened 2020-10-06
Support different list styles

PiotrKozlowski:


src/plugins/list/commands.js
```diff
@@ -14,6 +16,6 @@ export function changeListType(change, type) {
     if (node === null) {
         change.wrapInList(type)
     } else {
-        change.setNodeByKey(node.key, type)
```
aiwenar, line 17: Only properties related to list's type should be cleared. Aside from `class`, `mark-prefix`, `mark-suffix`, `item-sep`, and `type` should also be preserved.

src/cnxml/util.js
```diff
@@ -401,3 +401,12 @@ export const DEFAULT = {
         return mixedContent(el, next)
     },
 }
+
+export const cleanAttrs = obj => {
+    for (const key in obj) {
+        if (obj[key] === null || obj[key] === undefined || obj[key] === '') {
```
aiwenar, line 407: An empty string is a valid value (it's for example the default value for `item-sep`) so it shouldn't be filtered out, and remaining checks can be collapsed into `obj[key] == null`.

aiwenar (changes requested):

aiwenar (approved):

PiotrKozlowski:
@aiwenar I've squashed the fixup

PR #113, opened 2020-10-08
Support classes on document and handle language

PiotrKozlowski:
- support classes on document and glossary
- fix deserializing classes
- support language on the document
- do not allow empty classes

### DIFF
--- a/src/cnxml/de.ts
+++ b/src/cnxml/de.ts
@@ -85,6 +85,7 @@ export default function deserialize(
     }
 
     const doc: Doc = {
+        classes: Array.from(WithClasses.splitClasses(root.getAttribute('class') ?? '')),
         moduleId: root.getAttribute('module-id') ?? 'new',
         version: root.getAttribute('cnxml-version') as CnxmlVersion,
         title: '',
@@ -265,7 +266,7 @@ export function buildElement(
     }
 
     if (el.hasAttribute('class')) {
-        node.class = WithClasses.normalizeClasses([el.getAttribute('class')!])
+        node.classes = WithClasses.normalizeClasses([el.getAttribute('class')!])
     }
 
     for (const [key, value] of Object.entries(node)) {
--- a/src/cnxml/index.ts
+++ b/src/cnxml/index.ts
@@ -14,6 +14,8 @@ export type CnxmlVersion = '0.7' | '0.8'
 
 /** A serialized CNXML document */
 export interface Document {
+    /** Class names for the <document> element */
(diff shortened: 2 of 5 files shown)


src/cnxml/de.ts
```diff
@@ -85,6 +85,8 @@ export default function deserialize(
     }
 
     const doc: Doc = {
+        className: root.getAttribute('class') ?? undefined,
+        language: root.getAttributeNS('http://www.w3.org/XML/1998/namespace', 'lang') ?? undefined,
```
aiwenar, line 89: Language is already being deserialized in lines 98-100 (100-102 after your changes).

src/cnxml/index.ts
```diff
@@ -14,6 +14,8 @@ export type CnxmlVersion = '0.7' | '0.8'
 
 /** A serialized CNXML document */
 export interface Document {
+    /** Class name for the <document> element */
+    className?: string
```
aiwenar, line 18: While `Document` is not a Slate node, handling of classes on document should be compatible with `WithClasses`. This means that this field should be `classes: string[]`.

src/cnxml/se.tsx
```diff
@@ -64,11 +64,15 @@ export default function serialize(
         ? content.pop()!
         : null
 
+    const documentAttrs = doc.className ? { class: doc.className } : {}
```
aiwenar, line 67: You can just add `class={doc.className}` to JSX: this upgrade also includes a new XML serializer which filters out attributes whose `value == null`, so this trick is no longer needed.

src/interfaces/classes.ts
```diff
@@ -38,6 +38,6 @@ export const WithClasses = {
             result.push(...WithClasses.splitClasses(cls))
         }
 
-        return dedup(result.sort((a, b) => a.localeCompare(b)))
+        return dedup(result.filter(c => c.length).sort((a, b) => a.localeCompare(b)))
```
aiwenar, line 41: Empty classes should never be returned from `WithClasses.splitClasses`. I now see that they may be, but it's a bug (I'm too used to Python, where `' a  b '.split()` is `['a', 'b']` and not `['', 'a', '', 'b', '']`).

The proper fix would probably be to change `const WHITESPACE` to `/[ \t\n]+/`, and `splitClasses` to `yield* str.trim().split(WHITESPACE)`.

aiwenar (changes requested):

aiwenar (approved):

PR #117, opened 2021-02-03
Add support for Processing Instruction element

### LINKED ISSUE #119: Port #111 Add support for XML processing instructions in <content> to #102

PiotrKozlowski:
related to: https://github.com/openstax-poland/adaptarr-front/issues/512 

Fixes #119

### DIFF
--- a/src/cnxml/jsx.ts
+++ b/src/cnxml/jsx.ts
@@ -2,6 +2,8 @@
 // Licensed under the MIT license. See LICENSE file in the project root for
 // full license text.
 
+import { isPlainObject } from 'is-plain-object'
+
 import { Case } from '../interfaces'
 
 /** Qualified name of an element or attribute */
@@ -17,6 +19,7 @@ export interface Name {
     local: string
 }
 
+
 /** An element */
 export interface Element {
     /** Element's name */
@@ -29,8 +32,25 @@ export interface Element {
 
 export type Attributes = Omit<JSX.IntrinsicAttributes, 'children'> & { [key: string]: unknown }
 
+/** A processing instruction */
+export interface ProcessingInstruction {
+    /** PI's target */
+    target: string
+    /** PI's value */
+    value: string
+}
+
 /** Any value that can be used as child of a JSX element */
-export type Node = Element | globalThis.Node | string | Node[] | null
+export type Node = Element | ProcessingInstruction | globalThis.Node | string | Node[] | null
+
+export const Node = {
+    isElement(node: Node): node is Element {
+        return isPlainObject(node) && typeof (node as Element).name === 'object'
+    },
+    isProcessingInstruction(node: Node): node is ProcessingInstruction {
+        return node != null && typeof (node as ProcessingInstruction).target === 'string'
+    },
+}
 
 /** JSX element creator */
 export function createElement<
@@ -185,6 +205,8 @@ function finishElement(renderer: Renderer, element: Element, out: globalThis.Ele
             out.append(renderer.doc.importNode(child, true))
         } else if (typeof child === 'string') {
             out.append(child)
+        } else if (Node.isProcessingInstruction(child)) {
+            out.append(document.createProcessingInstruction(child.target, child.value))
         } else {
             out.append(renderElement(r, child))
         }
@@ -393,6 +415,11 @@ export declare namespace CNXML {
         value: string
     }
 
+    interface ProcesingInstruction extends Partial<Attributes.Common> {
+        target: string
+        value: string
+    }
+
     type Preformat = DisplayAny
     interface Problem extends Attributes.Common, Attributes.Typed {}
     interface Proof extends Attributes.Common, Attributes.Typed {}
@@ -488,6 +515,7 @@ export declare namespace JSX {
         note: IntrinsicAttributes & CNXML.Note
         para: IntrinsicAttributes & CNXML.Para
         param: IntrinsicAttributes & CNXML.Param
+        pi: IntrinsicAttributes & CNXML.ProcesingInstruction
         preformat: IntrinsicAttributes & CNXML.Preformat
         problem: IntrinsicAttributes & CNXML.Problem
         proof: IntrinsicAttributes & CNXML.Proof
--- a/src/cnxml/se.tsx
+++ b/src/cnxml/se.tsx
@@ -11,8 +11,8 @@ import {
     Admonition, AltText, Audio, Caption, Code, Commentary, CrossReference,
     Definition, DefinitionExample, DefinitionTerm, DocumentReference, Exercise,
     Figure, Footnote, Foreign, Glossary, Image, Link, List, ListItem, Meaning,
-    Media, MediaData, NumberStyle, Paragraph, Preformat, Problem, Proof,
-    Quotation, Rule, RuleExample, Section, SeeAlso, Solution, Statement,
+    Media, MediaData, NumberStyle, Paragraph, Preformat, Problem, ProcessingInstruction,
+    Proof, Quotation, Rule, RuleExample, Section, SeeAlso, Solution, Statement,
     StyledText, Term, Title, Video, WithClasses,
 } from '../interfaces'
 import { uuid } from '../util'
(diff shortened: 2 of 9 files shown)


src/interfaces/processingInstruction.ts
```diff
@@ -0,0 +1,15 @@
+import { Element } from 'slate'
+
+/** Processing Instruction */
+export interface ProcessingInstruction extends Element {
+    type: 'pi'
```
aiwenar, line 5: I would prefer a less ambiguous name, such as `processing_instruction`.

src/cnxml/jsx.ts
```diff
@@ -185,6 +186,11 @@ function finishElement(renderer: Renderer, element: Element, out: globalThis.Ele
             out.append(renderer.doc.importNode(child, true))
         } else if (typeof child === 'string') {
             out.append(child)
+        } else if (child.name.local === 'pi') {
+            out.append(document.createProcessingInstruction(
+                child.attributes.target as string,
+                child.attributes.value as string,
+            ))
```
aiwenar, line 193: This would prevent us from ever using an element named `pi`. It also confuses processing instructions with elements. Support for processing instructions should be more explicit:

1. Add new type for processing instructions:

   ```ts
   /** A processing instruction */
   export interface ProcessingInstruction {
      /** PI's target */
      target: string
      /** PI's value */
      value: string
   }
   ```

2. Add it as a possible `Node`:

   ```diff
    /** Any value that can be used as child of a JSX element */
   -export type Node = Element | globalThis.Node | string | Node[] | null
   +export type Node = Element | ProcessingInstruction | globalThis.Node | string | Node[] | null
   ```

3. Handle this new type in `renderChild`

   ```diff
    if (child instanceof globalThis.Node) {
       out.append(renderer.doc.importNode(child, true))
    } else if (typeof child === 'string') {
       out.append(child)
   +} else if (typeof child.target === 'string') {
   +   out.append(document.createProcessingInstruction(child.target, child.value))
    } else {
       out.append(renderElement(r, child))
    }
   ```

src/cnxml/se.tsx
```diff
@@ -496,6 +497,19 @@ function mediaItem(
...
+    node: ProcessingInstruction,
+    attrs: CommonAttrs,
+    children: JSX.Node,
+): JSX.Node {
+    return <pi
+        target={node.target}
+        value={node.value}
+        {...attrs}>
+        {children}
+    </pi>
```
aiwenar, line 510: Following my above comment about not confusing processing instructions with elements, this should be

```
return { target: node.target, value: node.value }
```

Regardless, adding `{...attrs}` and `{children}` here is confusing, as they will be ignored.

test/util/cnxml.js
```diff
@@ -3,7 +3,7 @@ export default global.cnxml = function cnxml(fragments) {
         throw new Error('cnxml`` does not support interpolation')
     }
 
-    return `<document xmlns="http://cnx.rice.edu/cnxml" cnxml-version="0.7" id="test" module-id="test">
+    return `<document xmlns="http://cnx.rice.edu/cnxml" cnxml-version="0.7" id="test" module-id="test" lang="en">
```
aiwenar, line 6: Why is this change made here?

aiwenar (changes requested):

test/util/cnxml.js
PiotrKozlowski: I've forgotten to remove this. I was trying to figure out why tests are failing

src/cnxml/jsx.ts
```diff
@@ -30,8 +30,28 @@ export interface Element {
...
+    node && typeof (node as ProcessingInstruction).target === 'string')
+
+export const nodeIsNotJSXElement = (
+    node: Node,
+// eslint-disable-next-line max-len
+): node is ProcessingInstruction | globalThis.Node | string | Node[] | null => typeof node === 'string'
+        || node === null
+        || Array.isArray(node)
+        || node instanceof globalThis.Node
+        || nodeIsProcessingInstruction(node)
```
aiwenar, line 54: Please use normal functions, and preferably a namespace:

```typescript
const Node = {
    isProcessingInstruction(...)

   ...
}
```

src/cnxml/se.tsx
```diff
@@ -401,14 +401,14 @@ function docref(node: DocumentReference, attrs: CommonAttrs, children: JSX.Node)
...
             return child.map(mapChild)
         }
 
+        if (JSX.nodeIsNotJSXElement(child)) {
+            return child
+        }
+
         if (child.name.namespace === JSX.CNXML_NAMESPACE
         && child.name.local === 'figure') {
             return {
```
aiwenar, line 414: This change is unrelated to the PR.

src/cnxml/se.tsx
```diff
@@ -440,12 +440,12 @@ function link(node: Link, attrs: CommonAttrs, children: JSX.Node): JSX.Node {
...
+
+        if (JSX.nodeIsNotJSXElement(child)) {
             return child
         }
 
-        if (Array.isArray(child)) return child.map(mapChild)
-
         if ((child.name.namespace == null || child.name.namespace === JSX.CNXML_NAMESPACE)
         && child.name.local === 'list') {
             return <item xmlns={JSX.CNXML_NAMESPACE}>{child}</item>
```
aiwenar, line 451: This change is unrelated to the PR.

aiwenar (changes requested):

src/cnxml/se.tsx
PiotrKozlowski: previous `if` statement was not including `ProcessingInstruction` so it was throwing an error since `child.name....` wasn't accessible on `child: Element | ProcessingInstruction`

src/cnxml/se.tsx
PiotrKozlowski: https://github.com/openstax-poland/cnx-designer/pull/117#discussion_r571933375

src/cnxml/jsx.ts
```diff
@@ -42,15 +42,15 @@ export interface ProcessingInstruction {
...
 export const Node = {
-    nodeIsNotJSXElement(node: Node):
+    isNotJSXElement(node: Node):
         node is ProcessingInstruction | globalThis.Node | string | Node[] | null {
         return typeof node === 'string'
             || node === null
             || Array.isArray(node)
             || node instanceof globalThis.Node
-            || Node.nodeIsProcessingInstruction(node)
+            || Node.isProcessingInstruction(node)
```
aiwenar: I would prefer a positive test (`isElement`). It will be simpler to read and won't need to be updated if we ever add another type of `Node`.

src/cnxml/se.tsx
aiwenar: I see. In this case please keep those checks in their original order when possible, this will make the diff smaller.

PiotrKozlowski: In this case, I thought that it would be simpler to create `nodeIsNotJSXElement` than creating an if statement with 4 branches and possible with new branches if we add new types to `Node`, but this required changing the order

src/cnxml/jsx.ts
```diff
@@ -29,8 +30,28 @@ export interface Element {
...
+
+export const Node = {
+    isElement(node: Node): node is Element {
+        return Boolean(node
+            && !Array.isArray(node)
+            && !Node.isProcessingInstruction(node)
+            && !(node instanceof globalThis.Node))
+    },
+    isProcessingInstruction(node: Node): node is ProcessingInstruction {
+        return Boolean(node && typeof (node as ProcessingInstruction).target === 'string')
```
aiwenar, line 52: Please don't use `Boolean(expression)` to check for truthness. If you want to check that `node` is not `null` do so with `node != null`.

`isElement` should check that `node` is an `Element` directly, rather than checking that it is _not_ any of the other possible types.

PiotrKozlowski: About the check for `isElement` - do you want me to check if the `node` is an object that contains `name` and that the name is an object that contains local of type string and that the node contains `attributes` of type object and also `children` that are of type `node` (which will require creating `Node.isNode` function)?

aiwenar: I don't think we need a full check, after all you are taking `node: Node`. I think something equivalent to `isPlainObject(node) && typeof node.name === 'object'` would suffice (using `isPlainObject` from package `is-plain-object`, it's already in our dependency tree as Slate uses it).

aiwenar (approved):

PiotrKozlowski:
In order to fix tests I've add `{ text: '' }` to `children` for
```ts
case Node.PROCESSING_INSTRUCTION_NODE:
            editor.apply({
                type: 'insert_node',
                path: path.current!,
                node: {
                    type: 'processing_instruction',
                    target: (child as ProcessingInstruction).target,
                    value: (child as ProcessingInstruction).data,
                    children: [{ text: '' }],
                },
            })
            break
```

Issue #116, opened 2021-02-03
Images are added without alt text field after slate update


jarosik10:
This issue exists in the updated version of the editor on branch upgrade.

PR #118, opened 2021-02-03
Add alt text insertnig media

jarosik10:
for: https://github.com/openstax-poland/cnx-designer/issues/116

### DIFF
--- a/src/normalize/figure.ts
+++ b/src/normalize/figure.ts
@@ -4,7 +4,7 @@
 
 import { Editor, Node, NodeEntry, Transforms } from 'slate'
 
-import { AltText, Caption, Figure, Media, Section } from '../interfaces'
+import { AltText, Audio, Caption, Figure, Image, Media, Section, Video } from '../interfaces'
 import { enumerate } from '../util'
 import { normalizeOrderedChildren } from './util'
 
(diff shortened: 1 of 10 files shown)


aiwenar (changes requested):
This change will only affect media elements created using `insertFigure`, but not ones created using `insertSubfigure` or manually via `Transforms.insertNodes`. It also won't ensure that all `Media` elements have alt texts, as `AltText` can still be removed after `Media` is created. 

It will also insert a second `AltText` if `media` argument to `insertFigure` already is a `Media` node with an `AltText` child.

Finally, requiring that all `Media` elements have `AltText` is part of document schema, which in Slate is enforced with normalizations. Please at this as a normalization in `src/normalizations/figure.ts`.

However, it might still be worth including `AltText` in newly created figures as it would then save one round of normalizations. In such case, please do it in `itemToChildren`.

jarosik10:
@aiwenar Do we want all `Media` elements include `alt`, even audio? This normalization makes some test fail right now.

src/normalize/figure.ts
```diff
@@ -87,6 +87,16 @@ export default function normalizeFigure(editor: Editor, entry: NodeEntry): boole
...
             return true
         }
+
+        // Media elements must contain alt text.
+        if (node.children.length > 0 && !AltText.isAltText(node.children[node.children.length - 1])) {
+            const alt = {
+                type: 'media_alt',
+                children: [{ text: '' }]
+            }
+            Transforms.insertNodes(editor, alt, { at: [...path, node.children.length] })
```
aiwenar, line 97:
```suggestion
            Transforms.insertNodes(editor, {
                type: 'media_alt',
                children: [{ text: '' }]
            }, { at: [...path, node.children.length] })
```

src/transforms/figure.ts
```diff
@@ -4,18 +4,26 @@
...
 export type MediaItem = Media | Omit<Audio | Image | Video, 'children'>
 
 function itemToChildren(media: MediaItem): Node[] {
+    const alt = {
+        type: 'media_alt',
+        children: [{ text: '' }]
+    }
+
     if (Media.isMedia(media)) {
+        if (!AltText.isAltText(media.children[media.children.length - 1])) {
```
aiwenar, line 18:
```suggestion
        if (!media.children.some(AltText.isAltText)) {
```

You shouldn't assume that existing `AltText` element will be last.

aiwenar (changes requested):
> Do we want all `Media` elements include `alt`, even audio? This normalization makes some test fail right now.

If you look at [CNXML documentation for the `<media>` element](https://legacy.cnx.org/eip-help/media) you'll notice that the `alt` attribute is required, regardless of what media elements there are. I'd say that tests need to be adjusted.

aiwenar (approved):
Looks good now. Please squash commits and it will be good to merge.

jarosik10:
@aiwenar should I fix the tests inside this PR as well? And how exactly should the commits be squashed?

aiwenar:
> should I fix the tests inside this PR as well?

Yes, these are all failures related to this PR.

> And how exactly should the commits be squashed?

Such that the commit history reflects evolution of the project, rather than the back and forth between the committer and the reviewer.

PiotrKozlowski:
@aiwenar Prośba o merge też tego pr jeżeli wszystko się zgadza

PR #120, opened 2021-02-09
Term name born died

PiotrKozlowski:
related to: https://github.com/openstax-poland/adaptarr-front/issues/515

### DIFF
--- a/src/cnxml/jsx.ts
+++ b/src/cnxml/jsx.ts
@@ -410,7 +410,10 @@ export declare namespace CNXML {
     interface Subfigure extends Attributes.Common, Attributes.Typed {}
     interface Sup extends Partial<Attributes.Common> {}
     interface Statement extends Attributes.Common, Attributes.Typed {}
-    type Term = Partial<Attributes.Common & Attributes.Link> & CMLNLE.Attributes.Reference & CXLXT.Attributes.Index
+    type Term = Partial<Attributes.Common & Attributes.Link>
+    & CMLNLE.Attributes.Reference
+    & CXLXT.Attributes.NameIndex
+    & CXLXT.Attributes.Index
     interface Title extends Partial<Attributes.Common> {}
     interface Video extends Partial<Attributes.Common>, Attributes.PlayableMedia, Attributes.VisualMedia {}
 }
@@ -432,6 +435,12 @@ export declare namespace CXLXT {
     namespace Attributes {
         interface Index {
             cxlxtIndex?: string
+        }
+
+        interface NameIndex {
+            cxlxtBorn?: number
+            cxlxtDied?: number
+            cxlxtName?: string
         }
     }
 }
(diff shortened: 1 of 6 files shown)


src/interfaces/glossary.ts
```diff
@@ -27,6 +27,9 @@ export interface Term extends Element {
     type: 'term'
     index?: string
     reference?: string
+    name?: string
+    born?: number
+    died?: number
```
aiwenar, line 32: Since these three attributes are only allowed when `index === "name"`, I'd rather express this as

```typescript
export interface NameTerm extends Term {
    index: 'name'
    name?: string
    born?: number
    died?: number
}
```

Although I'm not sure what would be a good name for this type.

src/cnxml/se.tsx
```diff
@@ -506,10 +506,19 @@ function rule(node: Rule, attrs: CommonAttrs, children: JSX.Node): JSX.Node {
...
+        nameIndexAttributes.cxlxtDied = node.died
+    }
+
     return <term
         xmlns={JSX.CNXML_NAMESPACE}
         cxlxtIndex={node.index}
         cmlnleReference={node.reference}
+        {...nameIndexAttributes}
         {...attrs}>
         { children }
```
aiwenar, line 523: While we only allow these attributes on terms whose `index === 'name'`, it is not enforced by the XML schema, and I'm not sure whether it's worth adding extra complexity to the exporter to omit them if `index` is different.

src/cnxml/jsx.ts
```diff
@@ -430,9 +435,21 @@ export declare namespace CMLNLE {
...
+            cxlxtDied?: number
+        }
+
         interface Index {
             cxlxtIndex?: string
         }
+
+        interface Name {
+            cxlxtName?: string
+        }
```
aiwenar, line 452: Since those attributes are meant to be used together, this would be better modelled by a single interface (say `Name`) containing all three.

aiwenar (changes requested):
Values of `born` and `died` are defined as `gYear` from [XMLSchema-datatypes](https://www.w3.org/TR/xmlschema-2/#gYear). Assuming I read it correctly it looks like they must always be at least four digits. We might want to implement custom formatting (I'm OK with not doing it in this PR).

src/cnxml/se.tsx
PiotrKozlowski: Since now they are split into separate interfaces I've had to either use `if` statement or cast the types to assigned these attributes

src/interfaces/glossary.ts
aiwenar: While I said that I'm not sure what would be a good name for this type, I think `TermForNameIndex` is a bit excessive :sweat_smile:.

src/cnxml/se.tsx
aiwenar: Ah, since `Term extends Element`, `Term.name: unknown`, which implies `(Term | TermForNameIndex).name: unknown | undefined | string`, which is the same as `undefined`. 

Right, in this case that if may be the best solution. I'll just ask that you change type of `nameIndexAttributes` to `JSX.CXLXT.Attributes.NameIndex` (and perhaps rename it to something less verbose).

PiotrKozlowski: done

aiwenar (approved):

PiotrKozlowski:
@aiwenar  Ready to merge

PiotrKozlowski:
I've resolved conflicts with `upgrade` branch

PR #123, opened 2021-03-05
Fix deserialization of empty inlines

PiotrKozlowski:
for: https://app.zenhub.com/workspaces/adaptarr-5d2f43c60f37561a03177ab7/issues/openstax-poland/adaptarr-front/548

### DIFF
--- a/src/cnxml/de.ts
+++ b/src/cnxml/de.ts
@@ -257,6 +257,10 @@ export function children(
         }
     }
 
+    if (el.childNodes.length === 0) {
+        editor.apply({ type: 'insert_node', path: path.current!, node: { text: '' } })
+    }
+
     path.unref()
 }
 
@@ -594,7 +598,7 @@ function link(editor: DeserializingEditor, el: Element, at: Path): void {
             target,
             document,
             case: el.getAttributeNS(CMLNLE_NAMESPACE, 'case'),
-            children: [{ text: '' }],
+            children: [],
         }, {})
         normalizeVoid(editor, at)
     } else if (url != null) {
--- a/test/cnxml/de/empty-inlines.js
+++ b/test/cnxml/de/empty-inlines.js
@@ -1,0 +1,29 @@
+/** @jsx h */
+/** @jsxFrag 'fragment' */
+
+// It should work when paragraph starts or ends with an empty inline
+export const input = cnxml`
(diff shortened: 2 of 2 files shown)


PiotrKozlowski:
We should also try another solution with adding empty text node to empty inlines

src/cnxml/de.ts
```diff
@@ -381,8 +384,8 @@ export function normalizeVoid(editor: DeserializingEditor, at: Path): void {
         normalizeWhiteSpace(editor, at)
     }
 
-    if (node.children.length === 1
-    && Text.isText(node.children[0]) && node.children[0].text === '') {
+    if (node.children.length > 0
+    && node.children.every((child: Slate.Node) => Text.isText(child) && child.text === '')) {
```
PiotrKozlowski, line 388: I think this could stay the same if we update line `600` from this diff to `children: [],` instead of `children: [{ text: '' }],` but I feel like checking all children instead of the only first one is just better

PiotrKozlowski:
@aiwenar I've pushed a change that we've talked about

src/cnxml/de.ts
aiwenar: This check is right after a call to `normalizeWhiteSpace`, so any sequence of `Text` nodes should already have been collapsed into a single `Text` node. This means that at this point the void _must_ have exactly one empty `Text` child. If it doesn't then it will be normalized, and deserialization will fail at line 146.

src/cnxml/de.ts
```diff
@@ -228,6 +228,9 @@ export function children(
     for (const child of el.childNodes) {
         switch (child.nodeType) {
         case Node.ELEMENT_NODE:
+            if (child.nodeType === Node.ELEMENT_NODE && child.childNodes.length === 0) {
+                (child as Element).append('')
+            }
```
aiwenar, line 233: 1. You are already in `case Node.ELEMENT_NODE`, so a check for `child.nodeType === Node.ELEMENT_NODE` is unnecessary;

2. Mutating DOM tree will likely be more expensive than mutating Slate document, so I'd prefer to avoid it. I suggest instead doing

   ```typescript
   if (el.childNodes.length === 0) {
       editor.apply({ type: 'insert_node', path: path.current!, node: { text: '' } })
   }
   ```

   after this loop

PR #124, opened 2021-03-18
Add tests for terms inside foreigns - formatted xml

PiotrKozlowski:
Deserializing this element:
```
<foreign xml:lang="en"><term xmlns:cxlxt="http://katalysteducation.org/cxlxt/1.0" cxlxt:index="foreign">term inside foreign - from not formatted xml</term></foreign>
```
works correctly, however, deserializing this element:
```
<foreign xml:lang="en">
  <term xmlns:cxlxt="http://katalysteducation.org/cxlxt/1.0" cxlxt:index="foreign">term inside foreign - from formatted xml</term>
</foreign>
```
fails with `Error: Cannot find a descendant at path ....`

Serialization works correctly.

Error is being thrown in line `132` in `whitespace.ts`:
```ts
131 Transforms.liftNodes(editor, {
132    at: Editor.range(editor, Path.next(end.path), nodePath.current!),
133    match: Text.isText,
134 })
```

Element at `Path.next(end.path)` just doesn't exists.

This function is hard to debug for me since so I've added a temporary solution in `de.ts`. Please advise what would be the best approach to solve it.


src/cnxml/de.ts
```diff
@@ -228,6 +228,17 @@ export function children(
     for (const child of el.childNodes) {
         switch (child.nodeType) {
         case Node.ELEMENT_NODE:
+            if (
```
PiotrKozlowski, line 231: this is temporary solution, please take a look at the description of the pr

PR #127, opened 2021-03-22
Add normalization for quotations

PiotrKozlowski:
for: https://github.com/openstax-poland/adaptarr-front/issues/513

### DIFF
--- a/src/normalize/quotation.ts
+++ b/src/normalize/quotation.ts
@@ -1,0 +1,52 @@
+// Copyright 2021 OpenStax Poland
+// Licensed under the MIT license. See LICENSE file in the project root for
+// full license text.
+
+import { Editor, Node, NodeEntry, Range, Transforms } from 'slate'
+import { List } from 'slate-lists'
+
+import { Paragraph, Quotation, Title } from '../interfaces'
+import { normalizeOrderedChildren } from './util'
+
+/**
+ * Normalize an quotation
+ *
+ * Return true if entry was normalized and false otherwise.
+ */
+export default function normalizeQuotation(editor: Editor, entry: NodeEntry): boolean {
+    const [node, path] = entry
+
+    if (Quotation.isQuotation(node)) {
+        // A quotation must not be empty.
+        if (Editor.isEmpty(editor, node)) {
+            Transforms.removeNodes(editor, { at: path })
+            return true
+        }
+
+        // Unwrap nested quotation if it is not selected
+        // and it's the only child of another quotation.
+        if (node.children.length === 1
+        && Quotation.isQuotation(node.children[0])
+        && (editor.selection == null || !Range.includes(editor.selection, path))) {
+            Transforms.unwrapNodes(editor, { at: [...path, 0] })
+            return true
+        }
+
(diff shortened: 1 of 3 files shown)


src/normalize/quotation.ts
```diff
@@ -0,0 +1,39 @@
...
+                || Paragraph.isParagraph(child)
+                || List.isList(child)
+                || Quotation.isQuotation(child)
+            ) {
+                index++
+                continue
+            }
+
+            Transforms.liftNodes(editor, { at: [...path, index] })
+            return true
```
aiwenar, line 34: 1. Remove empty quotations

   ```typescript
   // A quotation must not be empty.
   if (node.children.length === 0) {
       Transforms.removeNodes(editor, { at: path })
       return true
   }
   ```

2. I know eslint doesn't catch this, but please keep formatting of `if`s consistent.

3. Please use `normalizeOrderedChildren` instead of walking children manually. I know that right now there is no difference (since ensuring that `Title` must be the first child is done by `Title`'s normalizations), but when we add support for citations there might be, as we'll probably want quotation's source to always be last.

4. Unwrap nested quotation if it is not selected and it's the only child of another quotation.

src/normalize/quotation.ts
```diff
@@ -0,0 +1,39 @@
...
+/**
+ * Normalize an quotation
+ *
+ * Return true if entry was normalized and false otherwise.
+ */
+export default function normalizeQuotation(editor: Editor, entry: NodeEntry): boolean {
+    const [node, path] = entry
+
+    if (Quotation.isQuotation(node)) {
+        // Quotation's may contain only Titles, Paragraphs,
```
aiwenar, line 19:
```suggestion
        // Quotations may contain only Titles, Paragraphs,
```

aiwenar (changes requested):

src/normalize/quotation.ts
PiotrKozlowski: I added normalizations for 1 and 4 but it seems like tests for these cases are already there and they are passing without these normalization

src/normalize/quotation.ts
```diff
@@ -0,0 +1,53 @@
...
+            Transforms.removeNodes(editor, { at: path })
+            return true
+        }
+
+        // Unwrap nested quotation if it is not selected
+        // and it's the only child of another quotation.
+        if (node.children.length === 1
+            && Quotation.isQuotation(node.children[0])
+            && (editor.selection == null || !Range.includes(editor.selection, path))
+        ) {
```
aiwenar, line 31: Last nitpick: please keep style of `if` conditions consistent.

PiotrKozlowski:
```js
if (node.children.length === 1
&& Quotation.isQuotation(node.children[0])
&& (editor.selection == null || !Range.includes(editor.selection, path))
) {
```
?

aiwenar:
```typescript
if (node.children.length === 1
&& Quotation.isQuotation(node.children[0])
&& (editor.selection == null || !Range.includes(editor.selection, path))) {
```

aiwenar (approved):

Issue #90, opened 2020-04-09
Tests should have more descriptive names


PiotrKozlowski:
Tests files should export const `name` with a better description of what they are doing.
For ex. when a file has a name `shift-enter.js` it sometimes contains commands to test `double shift+enter`

aiwenar:
We've decided that we want to split these tests into several simpler (such that each one only presses a key once).

PR #132, opened 2021-03-25
Update tests names to be more adequate

PiotrKozlowski:
for: https://github.com/openstax-poland/cnx-designer/issues/90


aiwenar:
We've decided that we want to split these tests into several simpler (such that each one only presses a key once).

PR #134, opened 2021-04-01
Normalize white space at edges of unselected inlines

aiwenar:


PiotrKozlowski (changes requested):
It needs some tests and probably should be disabled for `<code>` elements

PiotrKozlowski (approved):

PR #137, opened 2021-04-15
Handle enter in lists

PiotrKozlowski:


PiotrKozlowski:
@aiwenar Do you know why `single-enter-at-the-start-unwraps-list-item.js` is failing?
It is hilarious because it should be the same as the `second` `break()` in `double-enter-not-at-the-start-unwraps-list-item.js`

PiotrKozlowski:
Handled in slate-lists https://github.com/openstax-poland/slate-lists/pull/4

PR #139, opened 2021-05-11
Make admonition a valid parent of figure

jarosik10:
for: https://github.com/openstax-poland/adaptarr-front/issues/572
It seems that it is already possible to put table into a admonition.

### DIFF
--- a/src/normalize/figure.ts
+++ b/src/normalize/figure.ts
@@ -4,7 +4,9 @@
 
 import { Editor, Node, NodeEntry, Transforms } from 'slate'
 
-import { AltText, Audio, Caption, Figure, Image, Media, Section, Video } from '../interfaces'
+import {
+    Admonition, AltText, Audio, Caption, Figure, Image, Media, Section, Video,
+} from '../interfaces'
 import { enumerate } from '../util'
 import { normalizeOrderedChildren } from './util'
 
@@ -17,11 +19,11 @@ export default function normalizeFigure(editor: Editor, entry: NodeEntry): boole
     const [node, path] = entry
 
     if (Figure.isFigure(node)) {
-        // Figure can only be a child of sections, the document, and other
-        // figures.
+        // Figure can only be a child of sections, admonitions, the document,
+        // and other figures.
         const [parent, parentPath] = Editor.parent(editor, path)
         if (!Section.isSection(parent) && !Figure.isFigure(parent)
-        && parentPath.length > 0) {
+            && !Admonition.isAdmonition(parent) && parentPath.length > 0) {
             Transforms.liftNodes(editor, { at: path })
             return true
         }
--- a/test/normalization/figure/in-admonition.js
+++ b/test/normalization/figure/in-admonition.js
@@ -1,0 +1,16 @@
+/** @jsx h */
+
+export const checkSelection = false
+
+export const input = <editor>
+    <note>
+        <figure>
+            <media>
+                <img src="figure.png" intendedUse="all"><text /></img>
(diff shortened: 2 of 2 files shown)


PiotrKozlowski (changes requested):
There are lint errors
Please also add a test for this

test/normalization/admonition/valid-children.js
```diff
@@ -0,0 +1,56 @@
...
+        <figure>
+            <media>
+                <img src="figure.png" intendedUse="all"><text /></img>
+                <mediaalt>valid children</mediaalt>
+            </media>
+        </figure>
+    </note>
+</editor>
+
+export const output = <editor>
```
PiotrKozlowski, line 34:
```suggestion
export const output = input
```

test/normalization/admonition/valid-children.js
```diff
@@ -0,0 +1,56 @@
+/** @jsx h */
```
PiotrKozlowski, line 1: I know that I said you can add test in `admonition` but now I think it would make more sense to have this test in `figure` - sth like `figure-valid-parents`
I'll mark @aiwenar To review since he is responsible for this repository

PiotrKozlowski (changes requested):

test/normalization/admonition/valid-children.js
aiwenar: I think this will be our first direct test for child-parent relationship, but I agree that it would fit better with figure tests, possibly as `figure/in-admonition.js` but I'm open to suggestions.

To be clear, this would be a test specifically for figure in admonition, and shouldn't test other elements.

src/normalize/figure.ts
```diff
@@ -17,11 +19,14 @@ export default function normalizeFigure(editor: Editor, entry: NodeEntry): boole
...
+        // Figure can only be a child of sections, admonitions, the document,
+        // and other figures.
         const [parent, parentPath] = Editor.parent(editor, path)
-        if (!Section.isSection(parent) && !Figure.isFigure(parent)
-        && parentPath.length > 0) {
+        if (!Section.isSection(parent)
+            && !Figure.isFigure(parent)
+            && !Admonition.isAdmonition(parent)
+            && parentPath.length > 0
+        ) {
```
aiwenar, line 29:
```suggestion
        if (!Section.isSection(parent) && !Figure.isFigure(parent)
        && !Admonition.isAdmonition(parent) && parentPath.length > 0) {
```

I know this may be a bit confusing, but me and @PiotrKozlowski are using slightly different styles, and I believe consistency within a single repository is important.

aiwenar (changes requested):

test/normalization/admonition/valid-children.js
jarosik10: @aiwenar 
I have a feeling that making this type of tests `figure/in-{parent}` will produce a lot of test files is `figure/*` directory and it will become a little bit messy. What do you think about creating a subfolder inside `figure` which will contain only child-parent relation test?

aiwenar: I don't think it's worth adding tests for every possible parent, especially since they are unrelated to this PR. We are interested in adding support for figures in admonitions, so only a test for this single case should be added.

PiotrKozlowski (approved):
There is one line error related to your change. I'm not sure if other errors should be fixed in this PR

aiwenar (approved):
None of lint issues remaining are related to this PR. They also seem to be caused by different version of linter, because this PR passes linting for me locally.

PR #141, opened 2021-08-12
Support for <equation>s

aiwenar:


src/normalize/equation.ts
```diff
@@ -0,0 +1,38 @@
...
+    const [node, path] = entry
+
+    if (Equation.isEquation(node)) {
+        // Equations must not be empty
+        if (node.children.length === 0) {
+            Transforms.removeNodes(editor, { at: path })
+            return true
+        }
+
+        // Rules must only have a single child
```
jarosik10, line 20: Rules?

src/plugin.ts
```diff
@@ -8,8 +8,9 @@ import normalizeNode from './normalize'
 import { isInline, isVoid } from './interfaces'
 
 /** A Slate editor augmented with support for editing CNX documents. */
-// eslint-disable-next-line @typescript-eslint/no-empty-interface
 export interface CnxEditor extends Editor {
+    /** Can {@code element} be used as content of an {@link Equation}? */
+    isEquationContent(element: Element): boolean
```
jarosik10, line 13: This comment doesn't seems really clear to me.
btw. `<equation>` can be directly followed be `<math>`, right?
```html
<equation>
   <math>...</math>
</equation>
```

src/normalize/equation.ts
```diff
@@ -0,0 +1,38 @@
...
+        if (node.children.length > 1) {
+            Transforms.splitNodes(editor, { at: [...path, node.children.length - 1] })
+            return true
+        }
+
+        // Equation can only contain elements from a predefined set and in
+        // a strict order.
+        if (!Element.isElement(node.children[0]) || !editor.isEquationContent(node.children[0])) {
```
jarosik10, line 28: should it be only checked for the equation's first child?

jarosik10 (changes requested):

src/normalize/equation.ts
aiwenar: The previous `if` enforces that equations only have one child.

jarosik10 (approved):
Looks good:) please squash the fixups

PR #143, opened 2021-09-22
Fix finding glossary

jarosik10:
for: https://github.com/openstax-poland/adaptarr-backlog/issues/99

### DIFF
--- a/src/transforms/glossary.ts
+++ b/src/transforms/glossary.ts
@@ -50,7 +50,7 @@ export function addGlossaryDefinition(
         let [glossary, glossaryPath] = Editor.above(editor, {
             at: path,
             match: Glossary.isGlossary,
-        })!
+        }) ?? []
 
         let index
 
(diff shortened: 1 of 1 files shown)


src/transforms/glossary.ts
```diff
@@ -50,11 +50,11 @@ export function addGlossaryDefinition(
         let [glossary, glossaryPath] = Editor.above(editor, {
             at: path,
             match: Glossary.isGlossary,
-        })!
+        }) ?? []
 
         let index
 
-        if (glossary == null) {
+        if (glossary == null || glossaryPath == null) {
```
aiwenar, line 57: This is not necessary. `Editor.above` returns `[Element, Path]`, so if `glossary == null` then also `glossaryPath == null`.

jarosik10: then i'll have to use non-null assertion in `glossaryPath!`

PR #144, opened 2021-09-23
Fix normalization when Texts are inside Media elements

aiwenar:


src/normalize/figure.ts
```diff
@@ -2,7 +2,7 @@
 // Licensed under the MIT license. See LICENSE file in the project root for
 // full license text.
 
-import { Editor, Node, NodeEntry, Transforms } from 'slate'
+import { Editor, Node, NodeEntry, Path, Text, Transforms } from 'slate'
```
jarosik10, line 5: You've imported `Path`, but never actually used it.

jarosik10 (changes requested):

jarosik10 (approved):

PR #149, opened 2021-10-20
Load all ids from editor on initial apply

jarosik10:


src/id.ts
```diff
@@ -102,3 +103,18 @@ function walk(node: Node, callback: (element: Element) => void) {
         for (const child of node.children) walk(child, callback)
     }
 }
+
+function setEditorIds(editor: Editor): Set<string> {
```
aiwenar, line 107: `collectIds` would be a more appropriate name

src/id.ts
```diff
@@ -23,7 +23,8 @@ function apply(apply: (op: Operation) => void, ed: Editor, op: Operation) {
     let ids = ID_MAP.get(ed)
 
     if (ids == null) {
-        ID_MAP.set(ed, ids = new Set())
+        ids = setEditorIds(ed)
+        ID_MAP.set(ed, ids)
```
aiwenar, line 27:
```suggestion
        ID_MAP.set(ed, ids = setEditorIds(ed))
```

src/id.ts
```diff
@@ -102,3 +103,18 @@ function walk(node: Node, callback: (element: Element) => void) {
...
+        if (Element.isElement(node)) {
+            if ('id' in node && typeof node.id === 'string') {
+                ids.push(node.id)
+            }
+
+            for (const child of node.children) readAllIds(child)
+        }
+    }
+    editor.children.forEach(node => readAllIds(node))
+    return new Set(ids)
```
aiwenar, line 119:
```suggestion
    return new Set(Array.from(Editor.nodes(editor, {
        match: node => Element.isElement(node) && 'id' in node && typeof node.id === 'string'
    }), ([node]) => node))
```

PR #151, opened 2021-10-26
Add the selected word after Definition for

OneByteTwoByte:


src/transforms/glossary.ts
```diff
@@ -27,11 +27,33 @@ export function addGlossaryDefinition(
...
+                    || elem.type === "title"
+                    || elem.type === "preformat"
+                    || elem.type === "table_entry"
+                    || elem.type === "definition_term")) {
+                        nodeContent = JSON.parse(JSON.stringify(elem.children))
+                        return true
+                    }
+                    return getToBottom(elem.children)
+                })
+            }
```
jarosik10, line 48: It might be better to use an array of validators. Something like: `const VALID_DEFINITION_TERM_ELEMENTS = [Paragraph.isParagraph, Title.isTitle...]`.

Why do you check for `elem.type === "definition_term"`?

`nodeContent = elem.children` should be fine

jarosik10 (changes requested):

src/transforms/glossary.ts
OneByteTwoByte: Why do you check for elem.type === "definition_term"? : in case someone wants to copy the definition - maybe someone will come up with such an idea :). Should I get rid of it?

nodeContent = elem.children does not work, just like [...elem.children], this object is not extensible

Array of validators is great idea, I will do it this way

src/interfaces/table.ts
```diff
@@ -0,0 +1,72 @@
...
+export interface TableEntry extends Element {
+    type: 'table_entry'
+}
+
+export const TableEntry = {
+    /** Check if value of unknown type is a collection of media elements */
+    isTableEntry(value: unknown): value is TableEntry {
+        return Element.isElement(value) && value.type === 'table_entry'
+    },
+}
```
aiwenar, line 72: Why is code for tables included in this PR?

src/transforms/glossary.ts
```diff
@@ -27,11 +55,35 @@ export function addGlossaryDefinition(
...
+                                } else {
+                                    nodeToInsert.push(elem)
+                                }
+                            })
+                        }
+                    }
+                    return getToBottom(elem.children as Descendant[])
+                })
+            }
+        }
```
aiwenar, line 79: This should be a utility function, declared outside `addGlossaryDefinition`

I also don't understand why this filtering is necessary. Ensuring valid document structure should be done in normalizations.

src/transforms/glossary.ts
```diff
@@ -27,11 +55,35 @@ export function addGlossaryDefinition(
...
         if (at == null) return
 
+        const fragment: Descendant[] = Editor.fragment(editor, at)
+        const nodeToInsert: Node[] = []
+        const getToBottom = (descendant: Descendant[]) => {
+            if (Array.isArray(descendant)) {
+                descendant.some((elem: Descendant) => {
+                    if (VALID_DEFINITION_TERM_ELEMENTS.some(v => v(elem))) {
+                        if (Array.isArray(elem.children)) {
+                            JSON.parse(JSON.stringify(elem.children)).forEach((elem: Element) => {
```
aiwenar, line 65: What is this conversion for?

```suggestion
                            for (const elem of elem.children) {
```

src/transforms/glossary.ts
```diff
@@ -27,11 +55,35 @@ export function addGlossaryDefinition(
 
         if (at == null) return
 
+        const fragment: Descendant[] = Editor.fragment(editor, at)
+        const nodeToInsert: Node[] = []
+        const getToBottom = (descendant: Descendant[]) => {
+            if (Array.isArray(descendant)) {
+                descendant.some((elem: Descendant) => {
```
aiwenar, line 62: What is this call to `some` doing?

src/transforms/glossary.ts
```diff
@@ -27,11 +55,35 @@ export function addGlossaryDefinition(
 
         if (at == null) return
 
+        const fragment: Descendant[] = Editor.fragment(editor, at)
+        const nodeToInsert: Node[] = []
+        const getToBottom = (descendant: Descendant[]) => {
+            if (Array.isArray(descendant)) {
```
aiwenar, line 61: `Editor.fragment` returns `Descendant[]`, so this check is unnecessary; it is always an array.

aiwenar (changes requested):

src/transforms/glossary.ts
OneByteTwoByte: About this check - 
I also thought that extra checking was unnecessary, but if it's not there then I have a bug:
`TypeError: Cannot read properties of undefined`

src/interfaces/table.ts
OneByteTwoByte: I needed an interface for a table. Make it a different PR?

src/transforms/glossary.ts
OneByteTwoByte: All this filtering is needed, because in some cases the text formatting should be preserved, in some cases not.

src/transforms/glossary.ts
OneByteTwoByte: If I do as you suggest, I get an error:
`TypeError: Cannot assign to read only property 'id' of object`

src/transforms/glossary.ts
aiwenar: I still don't understand why is it being done _here_, rather than in normalizations.

src/transforms/glossary.ts
aiwenar: Was that error by any chance in [src/id.ts:36](https://github.com/openstax-poland/cnx-designer/blob/master/src/id.ts#L36)?

src/transforms/glossary.ts
OneByteTwoByte: I will move that to normalizations

src/transforms/glossary.ts
OneByteTwoByte: I am not quite sure about that, there is the link to error message:
https://docs.google.com/document/d/1YLaUgrjJqzpPbS9PFvqLzA2h6wLpv6KEOUIxQH8p1q4/edit?usp=sharing

src/transforms/glossary.ts
OneByteTwoByte: I've come to the conclusion that people can have extreme ideas like making a definition with 2 paragraphs. Should I remove it?

src/interfaces/table.ts
aiwenar: We don't know yet how to handle tables, so it shouldn't be added at all (tables are currently implemented in a ‘temporary’ way in adaptarr-front).

src/transforms/glossary.ts
aiwenar: Unfortunately this tells me nothing. To debug this issue and see where it actually occurs open debug tools, go to debugger (sources in chrome), and select “pause on exceptions” (you may also need to select “pause on caught exceptions”), and then trigger this issue again. The debugger will pause, and you'll be able to check exactly where it is happening.

src/transforms/glossary.ts
aiwenar: You answered a different question. I'm asking not why you written this code, but rather what exactly is it doing. As far as I understand, this is just a loop, but if that's the case, why is it implemented using `Array#some` instead of a `for` statement?

OneByteTwoByte: You're right, I'll change it to for

src/normalize/glossary.ts
```diff
@@ -149,6 +149,21 @@ export default function normalizeGlossary(editor: Editor, entry: NodeEntry): boo
...
+                if (!Text.isText(child) && Node.children(editor, childPath)) {
+                    removeTerm(childPath)
+                }
+                if (Term.isTerm(child)) {
+                    Transforms.unwrapNodes(editor, { at: childPath })
+                    return true
+                }
+            }
+        }
+        removeTerm(path)
```
aiwenar, line 165: You shouldn't do it like this, since all children of `node` will be normalized. Instead in normalization for `Term` check, that it's not nested in a `DefinitionTerm`:

```ts
if (Editor.above(editor, { at: path, match: DefinitionTerm.isDefinitionTerm }) != null) {
    // is descendant of `DefinitionTerm`
}
```

src/transforms/glossary.ts
```diff
@@ -14,6 +20,7 @@ import { Definition, DefinitionExample, Glossary, Meaning, SeeAlso } from '../in
  * If select is set to true the selection will be collapsed into the new
  * definition.
  */
+
```
aiwenar, line 23: Remove this newline.

src/id.ts
```diff
@@ -32,6 +32,7 @@ function apply(apply: (op: Operation) => void, ed: Editor, op: Operation) {
     switch (op.type) {
     case 'insert_node':
         walk(op.node, el => {
+            el = { ...el }
```
aiwenar, line 35: Change unrelated to this PR.

src/normalize/glossary.ts
```diff
@@ -169,6 +184,11 @@ export default function normalizeGlossary(editor: Editor, entry: NodeEntry): boo
                 return true
             }
         }
+        const next = Path.next(path)
+        if (next) {
```
aiwenar, line 188: This condition is always true.

src/transforms/glossary.ts
```diff
@@ -27,11 +34,31 @@ export function addGlossaryDefinition(
...
+                for (const {} of path.slice(0, path.length-1)) {
+                    toInsert = toInsert[0].children
+                }
+            } else {
+                // eslint-disable-next-line no-empty-pattern
+                for (const {} of path) {
+                    toInsert = toInsert[0].children
+                }
+            }
+        }
```
aiwenar, line 55: This will discard any formatting which is around selection.

Instead do something like this

```ts
const fragment = unpackBlocks(editor, Editor.fragment(editor, at)

function unpackBlocks(editor: Editor, fragment: Descendant[]): Descendant[] {
    function* unpack(nodes: Descendant[]) {
        for (node in nodes) {
            if (Editor.isBlock(editor, node)) {
                yield* unpack(editor, node.children)
            } else {
                yield node
            }
        }
    }

    return Array.from(unpack(fragment))
}
```

src/transforms/glossary.ts
```diff
@@ -67,6 +94,7 @@ export function addGlossaryDefinition(
             at: [...glossaryPath!, index],
             select,
         })
+        return true
```
aiwenar, line 97: What is the purpose of this return?

aiwenar (changes requested):

OneByteTwoByte:
I have decided to stay with the way I did it before, because I had problems with the generator you proposed. I've just adjusted it, so the formatting is beeing kept now

src/normalize/glossary.ts
```diff
@@ -174,6 +167,13 @@ export default function normalizeGlossary(editor: Editor, entry: NodeEntry): boo
         }
     }
 
+    // We look for Term in DefinitionTerm, unwrap it if found
+    if (Term.isTerm(node) || Title.isTitle(node) || Paragraph.isParagraph(node)) {
```
aiwenar, line 171: You don't need to check for `Title` and `Paragraph`. These are block nodes, and so cannot be children of `DefinitionTerm`.

src/normalize/glossary.ts
```diff
@@ -185,10 +185,8 @@ export default function normalizeGlossary(editor: Editor, entry: NodeEntry): boo
             }
         }
         const next = Path.next(path)
-        if (next) {
-            const common = Path.common(path, next)
-            Transforms.mergeNodes(editor, { at: common })
-        }
+        const common = Path.common(path, next)
+        Transforms.mergeNodes(editor, { at: common })
```
aiwenar, line 189: This is incorrect. This will always merge `SeeAlso` with the next node.

src/normalize/glossary.ts
```diff
@@ -174,6 +167,13 @@ export default function normalizeGlossary(editor: Editor, entry: NodeEntry): boo
         }
     }
 
+    // We look for Term in DefinitionTerm, unwrap it if found
+    if (Term.isTerm(node) || Title.isTitle(node) || Paragraph.isParagraph(node)) {
+        if (Editor.above(editor, { at: path, match: DefinitionTerm.isDefinitionTerm }) != null) {
+            Transforms.unwrapNodes(editor, { at: path })
+        }
+    }
```
aiwenar, line 175: This should be grouped with normalizations for `Term` in `src/normalize/text.ts`.

aiwenar (changes requested):
> I had problems with the generator you proposed

Could you elaborate?

> I've just adjusted it, so the formatting is beeing kept now

The only change I see is `path.length-1` to `path.length-2`. Did you forget to commit, or am I missing something?

src/id.ts
OneByteTwoByte: The error I get without this change in some cases: 
TypeError: Cannot add property id, object is not extensible

przykryy (approved):

PR #153, opened 2021-11-30
added check for parent of media and wrap in figure in case it is not …

OneByteTwoByte:
…figure


src/normalize/figure.ts
```diff
@@ -85,6 +85,7 @@ export default function normalizeFigure(editor: Editor, entry: NodeEntry): boole
 
     if (Media.isMedia(node)) {
         // Ensure children are ordered media items then alt text.
+        const [parent] = Editor.parent(editor, path)
```
aiwenar, line 88: This line should be directly before the if using `parent`.

aiwenar (changes requested):

PR #154, opened 2021-11-30
pressing enter when cursor is at the end of the line in caption does …

OneByteTwoByte:
…not make an error

### DIFF
--- a/src/handlers.ts
+++ b/src/handlers.ts
@@ -136,7 +136,7 @@ function onEnter(editor: Editor, ev: KeyboardEvent): void {
                 Editor.deleteFragment(editor)
             }
 
-            Transforms.splitNodes(editor)
+            Transforms.splitNodes(editor, { always: true })
             Transforms.liftNodes(editor, { at: Path.next(captionPath!) })
         })
 
--- a/test/handlers/figure/enter-at-end-of-caption.js
+++ b/test/handlers/figure/enter-at-end-of-caption.js
@@ -1,0 +1,24 @@
+/** @jsx h */
+
+export default input => input.break()
+
+export const input = <editor>
+    <figure>
+        <media>
+            <img src="first.png" intendedUse="all"><text/></img>
+            <mediaalt>First picture</mediaalt>
+        </media>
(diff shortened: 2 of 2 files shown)


src/handlers.ts
```diff
@@ -129,15 +129,16 @@ function onEnter(editor: Editor, ev: KeyboardEvent): void {
...
     if (caption != null) {
         Editor.withoutNormalizing(editor, () => {
             if (!Range.isCollapsed(selection)) {
                 Editor.deleteFragment(editor)
             }
 
-            Transforms.splitNodes(editor)
-            Transforms.liftNodes(editor, { at: Path.next(captionPath!) })
+            if (!ev.shiftKey) {
+                Editor.insertText(editor, '\n')
```
aiwenar, line 140: A new line (`'\n'`) in this context will be normalized to a space, so this change effectively makes enter and space equivalent.

What the original code intended to do, when an enter is pressed inside a caption, is split the caption in two and move the part after the cursor out of the figure. The important part is moving out of figure, not the visual appearance of a break.

aiwenar (changes requested):

src/handlers.ts
```diff
@@ -135,8 +135,7 @@ function onEnter(editor: Editor, ev: KeyboardEvent): void {
             if (!Range.isCollapsed(selection)) {
                 Editor.deleteFragment(editor)
             }
-
```
aiwenar, line 138: This line was unnecessarily removed.

aiwenar (approved):
Looks good now.

In future please only force-push once a PR has been approved.

aiwenar:
It might also be good to add a test case that would crash without this fix.

OneByteTwoByte:
I have added test for this issue. It crashes without this fix. Is this ok?

aiwenar (approved):
> I have added test for this issue. It crashes without this fix. Is this ok?

That was the issue, wasn't it :wink: 

Aside from that, please rebase this PR such that both the fix and test are in a single commit, and it will be good to merge.

jarosik10 (approved):

PR #156, opened 2021-12-10
now title can be at the beginning of a draft

OneByteTwoByte:


src/normalize/structure.ts
```diff
@@ -24,6 +24,16 @@ export default function normalizeStructure(editor: Editor, entry: NodeEntry): bo
...
         }
+
+        // If the document starts with a title,
+        // the title must be wrapped in a section
+        if (Title.isTitle(editor.children[0])) {
+            Transforms.wrapNodes(editor, {
+                type: 'section',
+                children: [],
+            }, { at: [0] })
+            Transforms.mergeNodes(editor, { at: [1] })
```
aiwenar, line 35: `mergeNodes` is not appropriate here, as it is supposed to, well, _merge_ nodes. In this case it will move contents of `[1]` into the new section. That is not what we want here.

In case of elements like paragraphs, which contain text nodes, this will move raw text into the new section, triggering another normalization (and if I'm not mistaken, removing these text nodes entirely).

What you want to do here is 
```ts
Transforms.moveNodes(editor, { at: [1], to: [0, 1] })
```

aiwenar (changes requested):
I would also like to see a test case for this added.

aiwenar (approved):
Please squash fix commit (81b5065) into the main commit (b352115) and it'll be good to merge.

PR #170, opened 2023-05-10
id bug- link to figure problem, 926

OneByteTwoByte:


src/id.ts
```diff
@@ -43,6 +43,10 @@ function apply(apply: (op: Operation) => void, ed: IdEditor, op: Operation) {
             }
             add.push(el.id as string)
         })
+        if (!op.node.id) {
+            op.node.id = ed.generateID()
+            add.push(op.node.id as string)
+        }
```
aiwenar, line 49: This should be handled by the `walk` above. The reason it is not is that you made `walk` copy all nodes in c0b9a2c, so the assignment on line 42 doesn't affect the document.

aiwenar (changes requested):

aiwenar (approved):

